PR #3, opened 2015-06-15
Adding Dan's changes

wmarshall484:
Added Dan's changes and updated getting started guide.


chanskw:
Merging latest changes... so we are working off from the same base.

PR #30, opened 2015-06-24
Added support for including third party jars/classes.

wmarshall484:


java/src/com/ibm/streamsx/topology/Topology.java
```diff
@@ -398,6 +398,48 @@ public Topology topology() {
...
+     * 
+     * @param location The location of a jar to be included in the application's
+     * runtime when submitting with a DISTRIBUTED or STANDALONE context.
+     */
+    public void addThirdPartyDependency(String location) {
+       JavaFunctional.addThirdPartyDependency(this, location); 
+    }
+    
+    /**
+     * Includes a class file into the application's runtime. The following
```
ddebrunner: @wmarshall484 Isn't it doing more that including a class file? Namely it's including the complete set of classes that include the passed in class, which is the top-level 'classes; directory or the class's jar file.

java/src/com/ibm/streamsx/topology/builder/BOperatorInvocation.java
```diff
@@ -219,7 +219,10 @@ public JSONObject complete() {
         return json;
     }
 
-    // TEMP
+    // Needed by the DependencyResolver to determine whether the operator
```
ddebrunner: @wmarshall484 Is this comment valid?

java/src/com/ibm/streamsx/topology/Topology.java
```diff
@@ -398,6 +398,48 @@ public Topology topology() {
...
+     * Topology top = new Topology("myTopology");
+     * top.addThirdPartyDependency(myThirdPartyResource.class);
+     * </pre></code> 
+     * As with
+     * {@link com.ibm.streamsx.topology.Topology#addThirdPartyDependency(String location)},
+     * this is only required when using third-party libraries when running with
+     * a DISTRIBUTED or STANDALONE context.
+     * <br><br>
+     * If the .class file is part of a package structure, the package will be
+     * copied and compressed into a jar file to be included in the Applicatoin's
```
ddebrunner: @wmarshall484 typo on Applicatoin's

'package structure' - is this how the top-level structure (classes directory) is normally referred to? Could imply that only that class's package is included.

ddebrunner:
Looks good!

Just wondering about the name of the methods 'addThirdPartyDependency', given that a developer may want to add a jar that they created, thus it's not a third party's jar.

Maybe simpler to call the methods based upon what they do:

addJarDependency ?? - I know it can also be used for classes directory, but the primary use will be jars.
addDependency ?? - Maybe too generic.

addClasspathDependency - ??

PR #47, opened 2015-06-25
Fixes #45

### LINKED ISSUE #45: Add StreamsContext.isSupported(Topology); fix is-supported checking of "marker" ops
Following on from #22:
- Add StreamsContext.isSupported(Topology)
- fix EMBEDDED StreamsContext.submit()'s is-supported check to properly deal with (allow) "marker" ops:  union(),parallel(),unparallel()

dlaboss:
- Added StreamsContext.isSupported(Topology)
- Added new test SimpleEmbeddedTest
- fixed EMBEDDED StreamsContext.submit() is-supported check to allow
marker ops.

### DIFF
--- a/java/src/com/ibm/streamsx/topology/builder/GraphBuilder.java
+++ b/java/src/com/ibm/streamsx/topology/builder/GraphBuilder.java
@@ -160,14 +160,18 @@ public class GraphBuilder extends BJSONObject {
      */
     public void checkSupportsEmbeddedMode() throws IllegalStateException {
         for (BOperator op : ops) {
+            // note: runtime==null for markers
             String runtime = (String) op.json().get("runtime");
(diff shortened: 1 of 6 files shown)


ddebrunner:
At some point this code needs to move into the embedded content, for encapsulation. GraphBuilder should have no knowledge of context types.

dlaboss:
Dan, understood.  Would you have the embedded context call GraphBuilder.complete() and scan the JSON and know how to identify java and "marker" ops?  
Perhaps something like Map<String,Integer> GraphBuilder.contentSummary() would be better?  The keys would be the "runtime" kinds (e.g., "spl.java", "spl.cpp") as well as "marker", and the value is the number of instances.

ddebrunner:
test now

PR #52, opened 2015-06-26
Changed organization for gettingstarted. Added content.

wmarshall484:


wmarshall484:
New organization can be previewed here: http://wmarshall484.github.io/streamsx.topology/index.html

ddebrunner:
Looks much better!

In the getting started intro there is this:
_If you would like to install the Java Application API on your own machine to use or contribute to streamsx.topology, follow the instructions for cloning the Gitub repository and integrating it with eclipse or running via command line_

This seems out of place, suddenly jumping from quickstart to cloning the git repo. In addition I don't think there are instructions for cloning etc. So can we just keep these pages aimed at users of the api, so that sentence should be replaced with one that leads to the other sections (Eclipse and command line).

ddebrunner:
Command line instructions don't have any how to get the api, I think they can just be copied from the Eclipse section

PR #61, opened 2015-06-30
Add context submit support for job-name, job-group, etc

### LINKED ISSUE #60: Distributed submit: enable specifying job name and job group

dlaboss:
* Add JobProperties defining new properties
* Enhance Vwap sample to enable specifying context config properties
* Enhance distributed submit to support new job context properties and
existing ContextProperties.TRACING
* Enhance all invokers {build,standalone,distributed} to trace their
full cmdline

Fixes #60

### DIFF
--- a/samples/java/functional/src/simple/Util.java
+++ b/samples/java/functional/src/simple/Util.java
@@ -8,6 +8,7 @@ import java.io.File;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import com.ibm.streams.operator.logging.TraceLevel;
 import com.ibm.streamsx.topology.context.ContextProperties;
@@ -15,6 +16,7 @@ import com.ibm.streamsx.topology.context.JobProperties;
 import com.ibm.streamsx.topology.context.StreamsContext;
 
 public class Util {
+    private static Logger LOGGER = Logger.getLogger("simple.util");
     
     private static Map<String,Level> levelMap = new HashMap<String,Level>();
     static {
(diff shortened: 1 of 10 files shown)


samples/java/functional/src/vwap/Vwap.java
```diff
@@ -32,12 +37,15 @@
      * @param args
      */
     public static void main(String[] args) throws Exception {
+        String mode = args[0];
```
ddebrunner: The context type is a 'type', not a 'mode'

samples/java/functional/src/simple/Util.java
```diff
@@ -0,0 +1,95 @@
...
+                            value = toTracingLevel(valStr);
+                        else if (item.equals(ContextProperties.KEEP_ARTIFACTS))
+                            value = Boolean.valueOf(valStr);
+                        else if (item.equals(JobProperties.OVERRIDE_RESOURCE_LOAD_PROTECTION))
+                            value = Boolean.valueOf(valStr);
+                        else if (item.equals(JobProperties.PRELOAD_APPLICATION_BUNDLES))
+                            value = Boolean.valueOf(valStr);
+                        
+                        config.put(item, value);
+                        System.out.println("Setting config: item="+item+" value="+value+"  [arg="+arg+"]");
```
ddebrunner: This should be through a logger, not System.out. System.out is for the application output.

ddebrunner:
Do we need tests for this? E.g. verifying the data directory was set.

Code without tests will drop the code coverage numbers

dlaboss:
Yeah, test would be good but wasn't quite sure how to tackle checking all/some of these.
I suppose a test could capture logging output and observe the correct things were set in the streamtool submitjob request.  I'll look into it.

ddebrunner:
Added issue #65 to track adding tests, would prefer changes to be in the codeline.

Issue #70, opened 2015-07-02
Support .isolate(), .lowLatency(), and .endLowLatency() methods.


wmarshall484:
Currently, if a streams application is submitted with some form of a DISTRIBUTED context, the default is to run each operator in a separate PE. This is also the default for SPL programs, and has caused issues where customers create an excessive number of processes on their systems.

With the Java Application API, we are going to change the default behavior -- by default, every operator will run in the same PE, separated by threaded ports. 

If a Java Application API developers wants to separate subsequent operations into a separate PE they can call .isolate(), which returns a reference to a TStream whose child streams will run in a separate PE. To put this in SPL terms, everything runs in the same partition colocation by default; calling .isolate() creates another partition colocation in which the TStream's children run.

Lastly, introducing threaded ports between each operator incurs higher latency, so there will be a second mode of processing called "low latency". Low latency processing indicates that all operators will be created on the same PE <b>without</b> threaded ports between each other. Low latency processing can be started by calling a "lowLatency()" method on a TStream. This returns a reference to a TStream whose child operations will not be separated by threaded ports. Default processing may be resumed by calling an "endLowLatency()" method on a child TStream.

To reiterate: there are going to be three methods added to TStream:
.lowLatency()
.endLowLatency()
.isolate()

There will be two types of processing:
default - in which every operation (transformation, filtering, ...) is on the same PE, separated by threaded ports.
low latency - in which every operation is on the same PE, but are not separated by threaded ports.

hildrum:
If the threaded port is implemented in the C++-layer,  I think `lowLatency` and `endLowLatency` are probably not good names.  At least for large tuples, the most noticeable effect of sticking a threaded port before every operator is going to be higher CPU usage and lower overall throughput.   A user concerned about throughput may not consider a method with the name `lowLatency` is the opposite of what they want to do.
I think a name that implied merging or combining would be a better one--`merge` is already in use, but `consolidate`, `fuse`, `meld`, `compound`?

ddebrunner:
Doesn't it depend on what is downstream of the threaded port? With a complex function  (say image processing ) a threaded port is likely to increase throughput regardless of tuple size.

hildrum:
[...]

PR #85, opened 2015-07-08
Added support for .isolate() -- runs downstream operators in a separate PE.

wmarshall484:
Part of enhancement #70

### DIFF
--- a/java/src/com/ibm/streamsx/topology/TStream.java
+++ b/java/src/com/ibm/streamsx/topology/TStream.java
@@ -509,22 +509,30 @@ public interface TStream<T> extends TopologyElement {
     TStream<T> sample(double fraction);
     
     /**
-     * Add docs
-     * @return isolate tstream
+     * Return a TStream that runs in a separate process from the parent. The
+     * children of the returned TStream will also run in another process. If a
+     * child of the returned TStream is combined with another TStream through a 
+     * union(), both TStreams will run in another process, including their ancestors
+     * and children up until other isolate() boundaries.
+     * <br><br>
+     * In other words, for the following Topology:
+     * <pre><code>
+     *                    |--.isolate()---transform2--|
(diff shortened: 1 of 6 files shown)


wmarshall484:
test this

wmarshall484:
test this

wmarshall484:
test this

wmarshall484:
test this

java/src/com/ibm/streamsx/topology/TStream.java
```diff
@@ -507,6 +507,32 @@
...
+     * ------transform1---|                           |--.isolate()--filter---
+     *                    |--.isolate()---transform3--|
+     * </code></pre>
+     * 
+     * Three separate processes will be created -- one process will contain
+     * transform1, another will contain both transform2 and transform3, and the
+     * last will contain filter.
+     * 
+     * <br><br>
+     * Merging an isolated stream with its parent stream will throw an exception.
```
ddebrunner: What is "merging" ?

java/src/com/ibm/streamsx/topology/TStream.java
```diff
@@ -507,6 +507,32 @@
      *             than 1.0.
      */
     TStream<T> sample(double fraction);
+    
+    /**
+     * Return a TStream that runs in a separate process from the parent. The
```
ddebrunner: parent/child hasn't been used to describe stream relationships, so it's not clear what defines a parent-child relationship. I'm not sure we need to define such a relationship either.

I think all that isolate() defines is that any processing applied to the returned stream will be in a separate process to the processing that produced this stream.

Where there are no specific instructions from the application, the Java Application API is free to decide the layout of a topology to PEs, thus it's not guaranteed that a stream's "children" will be in the the same PE.

ddebrunner: Good to comment that this only applies for distributed.

Issue #73, opened 2015-07-05
Can I include files in the application bundle?


hildrum:
I didn't see this in the documentation.

Is there a way to include files and directories in the application bundle?  If so, how do I create a path to a file that is in the bundle?  Is there something like `getApplicationDir()`?

ddebrunner:
From Java code the natural way would be to include the files in your jar file or classes directory containing the code that builds the topology or include an additional jar/classes directory using Topology.addJarDependency() or Topology.addClassDependency().

Then you access them as resources in Java from the class loader:
https://docs.oracle.com/javase/tutorial/deployment/webstart/retrievingResources.html

hildrum:
Thank you.   

For anyone else trying to include a file in the jar to be accessed at runtime:
* Put the files you want to access in the `bin` directory if using StreamsStudio.  The default is not to include the source directories in the jar, so putting files under `src` doesn't get them included.
* If you need to use the path in a static context (as I did), you can do `ProcessLines.class.getClassLoader()` (where `ProcessLines` is the name of the class).  The example linked to above uses `this`, but you can't do that in a static context.

ddebrunner:
Putting source files under bin might be risky, I'm not sure if a clean project will delete the contents of bin. Eclipse might have a way to copy resources from source directories to bin.

dlaboss:
fwiw, Topology.addFileDependency() is being added as part of the Pull #103 / Issue #98 work.  Don't know if it may be applicable to this use case.

ddebrunner:
Files can be added as Java resources or using addFileDependency() for SPL operators.

PR #103, opened 2015-07-13
Add KafkaStreams source and sink support.

### LINKED ISSUE #98: Support reading messages from Kafka as a source stream
@dlaboss is working on this.

dlaboss:
Fixes #98
Add KafkaStreams producer/consumer adapters.
Add Topology.addFileDependency().


ddebrunner:
skip

ddebrunner:
That skip was for Jenkins!

java/src/com/ibm/streamsx/topology/kafka/KafkaSchemas.java
```diff
@@ -0,0 +1,27 @@
...
+# Copyright IBM Corp. 2015  
+ */
+package com.ibm.streamsx.topology.kafka;
+
+import static com.ibm.streams.operator.Type.Factory.getStreamSchema;
+
+import com.ibm.streams.operator.StreamSchema;
+
+/**
+ * SPL schema for tuples associated with streamsx.messaging
```
ddebrunner: Use the toolkit name, not the github project name

java/src/com/ibm/streamsx/topology/kafka/KafkaStreams.java
```diff
@@ -0,0 +1,284 @@
...
+                list.add(toKafkaTupleStream(rawKafka));
+            }
+            TStream<KafkaMessage> kafka = list.remove(0);
+            
+            return kafka = kafka.union(new HashSet<>(list));
+        }
+    }
+    
+    /**
+     * Same as {@code source(te, topic, 1, kafkaConsumerConfig)}
```
ddebrunner: Have the summary describe what it does. In the body you can say it is equivalent to another method.

java/src/com/ibm/streamsx/topology/kafka/KafkaStreams.java
```diff
@@ -0,0 +1,284 @@
...
+     * <li><code>request.required.acks</code></li>
+     * </ul>
+     * @param topic optional Kafka topic to publish to.
+     *          Whens specified, a {@link KafkaMessageImpl#getKafkaTopic()} value is ignored.
+     *          If null, {@code toKafkaTupleFunc} must generate tuples that have a topic.
+     * @param stream the stream to publish to Kafka
+     * @param toKafkaMessageFunc function to generate a {@code KafkaTuple} from a tuple {@code T}.
+     *          If {@code topic} is null, the generated {@code KafkaTuple} <b>must</b> contain a {@code topic}.
+     */
+    public static <T> void sink(TopologyElement te,
```
ddebrunner: Should the name be the kafka terminology, producer, not sink?

java/src/com/ibm/streamsx/topology/kafka/KafkaMessageImpl.java
```diff
@@ -0,0 +1,119 @@
...
+package com.ibm.streamsx.topology.kafka;
+
+import java.io.Serializable;
+
+/**
+ * Stream object class usable with KafkaStreams sources.
+ * 
+ * @see KafkaStreams
+ */
+public class KafkaMessageImpl implements Serializable, KafkaMessage {
```
ddebrunner: Is there any benefit to having a impl/interface split, it seems that the expectation is that developers would use KafkaMessageImpl.

If a split is needed, then maybe a better name, I tend to assume Impl classes are not part of the public api.

java/src/com/ibm/streamsx/topology/kafka/KafkaMessageImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public String toString() {
+        return "{topic=" + topic + ", key=" + key + ", message=" + message + "}";
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + message != null ? message.hashCode() : 0;
```
ddebrunner: how can message be null?

java/src/com/ibm/streamsx/topology/kafka/KafkaStreams.java
```diff
@@ -0,0 +1,284 @@
...
+     * <li><code>auto.commit.interval.ms</code></li>
+     * </ul>
+     * @param threadsPerTopic number of threads to allocate to processing each topic
+     * @param topics one or more Kafka topics to subscribe to
+     * @return TStream&lt;KafkaTuple>
+     *      The generated tuples do not have a {@code topic}.
+     * @throws IllegalArgumentException if topics is null or empty.
+     * @throws IllegalArgumentException if threadsPerTopic <= 0.
+     */
+    public static TStream<KafkaMessage> source(TopologyElement te, Properties kafkaConsumerConfig,
```
ddebrunner: consumer as the method name, instead of source?

java/src/com/ibm/streamsx/topology/kafka/KafkaStreams.java
```diff
@@ -0,0 +1,284 @@
...
+     *          If null, {@code toKafkaTupleFunc} must generate tuples that have a topic.
+     * @param stream the stream to publish to Kafka
+     * @param toKafkaMessageFunc function to generate a {@code KafkaTuple} from a tuple {@code T}.
+     *          If {@code topic} is null, the generated {@code KafkaTuple} <b>must</b> contain a {@code topic}.
+     */
+    public static <T> void sink(TopologyElement te,
+            Properties kafkaProducerConfig,
+            final String topic,
+            TStream<T> stream,
+            final Function<T,? extends KafkaMessage> toKafkaMessageFunc)
```
ddebrunner: There's an asymmetry in that TStream<KafkaMessage> is produced but the consumer does not take TStream<KafkaMessage>. Instead it takes any TStream<T> but introduces a new way to do tuple transformation.

An existing pattern would be to accept TStream<T extends KafkaMessage> and expect that TStream.transform() would be used for conversion. This would require the interface for KafkaMessage

Or maybe follow the pattern of JSON, where classes can implement an interface to convert to KafkaMessage (KafkaMessageable (really bad name)).

Not saying what is done is wrong, but just trying to think why a different approach was taken, and will we end up with many similar methods for all different types of output/input classes, thus transformation of tuples from one type to another is no longer a single building block?

Not sure

java/src/com/ibm/streamsx/topology/kafka/KafkaMessage.java
```diff
@@ -0,0 +1,28 @@
+package com.ibm.streamsx.topology.kafka;
+
+import java.io.Serializable;
+
+/**
+ * The components of an Apache Kafka Message.
+ */
+public interface KafkaMessage extends Serializable {
```
ddebrunner: A different approach, would be to support Kafka messages as JSONObject, which has strong existing interchange capabilities with SPL. Would neatly support the optional key and topic.

java/src/com/ibm/streamsx/topology/kafka/KafkaStreams.java
dlaboss: Good question.  Originally I had "producer" and "consumer" but found it rather confusing (in usage in the test code).  I switched to "sink" and "source" to make it less confusing.  "sink" and "source" are the Java API methods that would have been called/used by a user had I been able to instead offer up the functionality as a Java API "Supplier" and a "Consumer" implementation respectively.  And of course, the "Consumer" implementation would really be a kafka-producer. :-)
Happy to change it to whatever you want.  Just say the word.

java/src/com/ibm/streamsx/topology/kafka/KafkaMessageImpl.java
dlaboss: it can't.  but in this case (and similarly in equals()) keeping the same code pattern for the member handling didn't seem like a bad idea. i'll remove the those checks if you like.

hildrum:
What happens if you do Topology.addFileDependency() and give it a path that is not included in the application bundle by default--I'd guess it gets copied to the toolkit root, but doesn't end up in the bundle.  If so, would it make sense to add a message about that in the javadoc?

If I understand correctly, addFileDependency() puts the file in a location that can be used by SPL operators, but there doesn't seem to be a way for `Function`s to access it.

dlaboss:
To clarify, addFileDependency() restricts addition of artifacts to the "etc" and "opt" toolkit (hence bundle) directories.  I take it from Dan's comment in issue #73 that there's a more appropriate way for Function implementations to access resources at runtime.

ddebrunner:
jenkins test this please

ddebrunner:
the jenkins build keeps failing as it can't find the samples when running the tests. Nothing obvious in the changes jumps out at me, but there are changes to the test/build.xml.

Do the tests run for you, it's failing on this test:
Test com.ibm.streamsx.topology.test.samples.BundleSamplesTest failed

dlaboss:
jenkins test this please

java/src/com/ibm/streamsx/topology/Topology.java
```diff
@@ -502,6 +502,34 @@ public void addJarDependency(String location) {
...
+     * // add "myApp" directory tree to the bundle's "etc" directory
+     * addFileDependency("etc", "/tmp/myApp");
+     * </pre>
+     * 
+     * @param dstDirName name of directory in the bundle
+     * @param location path to a file or directory
+     * @throws IllegalArgumentException if {@code dstDirName} is not {@code etc}
+     *     or {@code opt}, or {@code location} is not a file or directory.
+     */
+    public void addFileDependency(String dstDirName, String location) {
```
ddebrunner: Isn't it typical to have the order be source then destination, e.g. linux command line copy, Java Files.copy etc.

Need to state that these are (for SPL apps) etc/opt in the application directory.

java/src/com/ibm/streamsx/topology/tuple/SimpleMessage.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param key
+     *            Optional key associated with the message.  May be null.
+     * @param topic
+     *            Kafka topic identifier.
+     * @throws IllegalArgumentException if topic == null
+     * @throws IllegalArgumentException if message == null
+     */
+    public SimpleMessage(String message, String key, String topic) {
+        if (topic == null)
+            throw new IllegalArgumentException("topic==null");
```
ddebrunner: Seems inconsistent that topic is checked for null but key never is, but both by the definition are optional.

java/src/com/ibm/streamsx/topology/messaging/kafka/KafkaSchemas.java
```diff
@@ -0,0 +1,28 @@
...
+    private KafkaSchemas() {
+    }
+
+    /**
+     * SPL schema used for {@link com.ibm.streamsx.topology.spl.SPLStream SPLStream}
+     * of Kafka messages.
+     * Consists of three attributes: {@code rstring topic}, {@code rstring key}
+     * and {@code rstring message}.
+     */
+    public static StreamSchema KAFKA = getStreamSchema(
```
ddebrunner: These seems to be a generic message schema, rather than specific to kafka

ddebrunner:
Good stuff!

I see there's a ConsumerConnector and a ProducerConnector, would it make more sense just to have a single object that is a Broker, that I can send or receiver messages from? Just seems if I'm using a single Kafka broker for getting and sending messages, I have to define the configuration twice.

java/src/com/ibm/streamsx/topology/messaging/kafka/package-info.java
```diff
@@ -0,0 +1,11 @@
+/*
+# Licensed Materials - Property of IBM
+# Copyright IBM Corp. 2015  
+ */
+/**
+ * 
+ * Apache Kafka related streams.
```
ddebrunner: Is more information needed here, some overview?

java/src/com/ibm/streamsx/topology/messaging/kafka/ConsumerConnector.java
```diff
@@ -0,0 +1,198 @@
...
+
+            @Override
+            public Message apply(Tuple tuple) {
+                return new SimpleMessage(fromSplValue(tuple.getString("message")),
+                                    fromSplValue(tuple.getString("key")),
+                                    fromSplValue(tuple.getString("topic")));
+            }
+            
+            private String fromSplValue(String s) {
+                // SPL doesn't allow null value.  For our use,
```
ddebrunner: Maybe it's too early and I'm not reading the code correctly, but doesn't this mean that an empty message (message attribute of "") will cause an exception? I would think that the code should just honour the message as having empty text.

Similarly with the code as it stands, and empty topic attribute will cause an exception, though see the other comment.

java/src/com/ibm/streamsx/topology/Topology.java
dlaboss: will do.

java/src/com/ibm/streamsx/topology/messaging/kafka/ConsumerConnector.java
ddebrunner: I think what ever the decision is made as to how to handle empty values, it should be clearly documented in the visible javadoc.

java/src/com/ibm/streamsx/topology/tuple/SimpleMessage.java
dlaboss: The idea was that the two param ctor was for the null-topic case and this variant was strictly for the non-null case.  But in practice that enforcement did make use a bit cumbersome... probably w/o any real benefit.  I'll remove the check.  That said, since I'm going back in, I think I'll toss the  and 3 (String) param ctors and add a Builder.

ddebrunner: Why not just constructors?

java/src/com/ibm/streamsx/topology/messaging/kafka/KafkaSchemas.java
dlaboss: Yup/OK.

dlaboss:
There are couple of parts to this:
- a kafka consumer and producer have different configuration properties, both just a generic collection of Properties.  They don't even share the same "server location" property.
- I imagine there will (typically?) be cases where a topology is either a consumer or a producer but not both so having the separate classes so the partitioning seemed OK.
- Having the separate classes/ctors ~enforced that one had to provide a producer/consumer config in order to get a connector where publish()/subscribe() can be called.

java/src/com/ibm/streamsx/topology/tuple/SimpleMessage.java
dlaboss: Easy to misorder message,key,topic as they're all String.  But OK. I'll stick w/them.

java/src/com/ibm/streamsx/topology/messaging/kafka/ConsumerConnector.java
dlaboss: good catch on the empty msg. thx. i've also added such a msg to one of the test cases.
I'll update the subscribe() and publish() doc to make the behavior clear.
Note, the topic can't ever be empty in a received message so that shouldn't have been an issue.

Sadly I don't seem how to keep this aspect of SPL and the handling of the optional key from leaking through to the interface.  Fortunately, an empty key seems about as useful as no key so presenting it as no-key (null) seems ok.

PR #106, opened 2015-07-14
Added support and tests for lowLatency, endLowLatency, and threaded port generation.

### LINKED ISSUE #70: Support .isolate(), .lowLatency(), and .endLowLatency() methods.

### LINKED ISSUE #96: Address TStream.isolate feedback
Address comments in push request #85

wmarshall484:
Fixes #96 
Fixes #70

### DIFF
--- a/java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
@@ -1,0 +1,286 @@
+package com.ibm.streamsx.topology.generator.spl;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+import com.ibm.json.java.JSONArray;
+import com.ibm.json.java.JSONObject;
+import com.ibm.streamsx.topology.function7.Consumer;
+
+public class GraphUtilities {
+    public static ArrayList<JSONObject> findStarts(JSONObject graph) {
+        ArrayList<JSONObject> starts = new ArrayList<JSONObject>();
+        JSONArray ops = (JSONArray) graph.get("operators");
+        for (int k = 0; k < ops.size(); k++) {
+            JSONObject op = (JSONObject) (ops.get(k));
+            JSONArray inputs = (JSONArray) op.get("inputs");
+            if (inputs == null || inputs.size() == 0) {
+                if (((JSONObject) ops.get(k)).get("name") != null
+                        && !((String) ((JSONObject) ops.get(k)).get("name"))
+                                .startsWith("$")) {
+                    starts.add((JSONObject) ops.get(k));
+                }
+            }
+        }
+        return starts;
+    }
+    
+    public static ArrayList<JSONObject> findOperatorByKind(String opKind, JSONObject graph) {
+        ArrayList<JSONObject> lowLatencyStarts = new ArrayList<JSONObject>();
+        JSONArray ops = (JSONArray) graph.get("operators");
+        for (int k = 0; k < ops.size(); k++) {
+            JSONObject op = (JSONObject) (ops.get(k));
+            String kind = (String) op.get("kind");
+            if (kind != null && kind.equals(opKind)) {
+                lowLatencyStarts.add(op);
+            }
+        }
+        return lowLatencyStarts;
+    }
+    
+    // Get the children of the operator. Should probably be re-worked.
+    public static ArrayList<JSONObject> getChildren(JSONObject visitOp,
+            JSONObject graph) {
+        ArrayList<JSONObject> uniqueChildren = new ArrayList<JSONObject>();
+        HashSet<JSONObject> children = new HashSet<JSONObject>();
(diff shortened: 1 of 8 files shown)


wmarshall484:
jenkins test this please

wmarshall484:
jenkins test this please

java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -0,0 +1,310 @@
...
+                    starts.add((JSONObject) ops.get(k));
+                }
+            }
+        }
+        return starts;
+    }
+
+    public static ArrayList<JSONObject> findOperatorByKind(String opKind,
+            JSONObject graph) {
+        ArrayList<JSONObject> lowLatencyStarts = new ArrayList<JSONObject>();
```
ddebrunner: Seems the method is generic so the collection should be called something like kindOperators, not lowLatencyStarts

java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -0,0 +1,310 @@
...
+            JSONObject op = (JSONObject) (ops.get(k));
+            String kind = (String) op.get("kind");
+            if (kind != null && kind.equals(opKind)) {
+                lowLatencyStarts.add(op);
+            }
+        }
+        return lowLatencyStarts;
+    }
+
+    // Get the children of the operator. Should probably be re-worked.
```
ddebrunner: Need some more detailed comments on what is returned, all downstream operators or only immediate? Any special handling for marker operators? Same for getParent.

Maybe consider chnaging the names to up/downstream.

test/java/src/com/ibm/streamsx/topology/test/api/IsolateTest.java
```diff
@@ -21,54 +23,87 @@
...
-        
-	// Jenkins seems to fail when running this with a distributed tester.
-	// Not sure why, but the tests pass on my local build. 
-        /*try{
-        StreamsContextFactory
-                .getStreamsContext(StreamsContext.Type.DISTRIBUTED_TESTER)
-                .submit(topology).get(90, TimeUnit.SECONDS);
-        } catch(Exception e){
+
+        // Jenkins seems to fail when running this with a distributed tester.
```
ddebrunner: The issue is that the test needs to follow the standard pattern of extending TestTopology and then have 'assume' checks that make it only run in distributed. currently Jenkins is not running distributed as it doesn't yet have an instance setup.

test/java/src/com/ibm/streamsx/topology/test/api/LowLatencyTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    
+    @SuppressWarnings("serial")
+    private static Function<String, String> getPEId(){
+        return new Function<String, String>(){
+            int counter = 0;
+            @Override
+                public String apply(String v) {
+                // TODO Auto-generated method stub
+                return ((BigInteger) PERuntime.getCurrentContext().getPE().getPEId()).toString();
```
ddebrunner: Curious why a cast to BigInteger is needed. Also there's an unused count and a TODO comment.

test/java/src/com/ibm/streamsx/topology/test/api/IsolateTest.java
```diff
@@ -21,54 +23,87 @@
...
+        TStream<String> ss5 = ss4.transform(getPEId(), String.class).isolate();
+        ss5.transform(getPEId(), String.class).print();
+
+        TStream<String> ss7 = ss3.transform(getPEId(), String.class);
+
+        StreamsContextFactory.getStreamsContext(StreamsContext.Type.TOOLKIT)
+                .submit(topology).get();
     }
 
     @Test(expected = IllegalStateException.class)
```
ddebrunner: Some comments as to why this should throw an exception, it looks valid to me, or is it a case that is just not yet covered?

PR #123, opened 2015-07-23
Fix improperly fused kafka connector operators.

dlaboss:
Continue to use SPL.invoke to avoid introducing a compile-time
dependency on streamsx.messaging jar. But identify the added ops as
"java" so they get the same "vmArgs" treatment as added JavaPrimitive
ops, avoiding the sc fusion constraints.

Add new TSink as a result from TStream.sink(), SPL.invokeSink() and
JavaPrimitive.invokeJavaPrimitiveSink() to enable getting a sink's
underlying operator. 

Fix the name of JavaPrimitive.invokeJavaSource() to match the convention
of the other JavaPrimitive.invoke*() methods.


java/src/com/ibm/streamsx/topology/TSink.java
```diff
@@ -0,0 +1,21 @@
+/*
+# Licensed Materials - Property of IBM
+# Copyright IBM Corp. 2015  
+ */
+package com.ibm.streamsx.topology;
+
+import com.ibm.streamsx.topology.builder.BOperatorInvocation;
+
+/**
+ * A {@link TopologyElement} handle to a {@code sink} operation.
```
ddebrunner: Really it's a handle to the sink of a TStream.

dlaboss:
Don't merge yet.  Something is mucked up and unittest.kafka is failing.  I have to investigate.

dlaboss:
as noted in the most recent commit (9be5950) I think we should merge these changes.  There are still a couple of outstanding problems but these changes make the bulk of the kafka integration / tests work.

ddebrunner:
jenkins test this please

PR #124, opened 2015-07-23
Add TStream.split() issue113

dlaboss:


java/src/com/ibm/streamsx/topology/TStream.java
```diff
@@ -89,6 +90,41 @@
...
+     * For each tuple on the stream, {@code splitter.applyAsInt(tuple)} is called.
+     * The return value {@code r} determines the destination stream:
+     * <pre>
+     * if r < 0 the tuple is discarded
+     * else it is sent to the stream at position (r % n) in the returned array.
+     * </pre>
+     * </P>
+     * 
+     * <P>
+     * Example of round robin splitting a stream of {@code String} into 5 channels.
```
ddebrunner: Since split is a logical split, can we have suitable example, e.g. based upon a type of field in the object. Also if a comment could be added as to what makes it different to parallel().

java/src/com/ibm/streamsx/topology/internal/functional/ops/FunctionSplit.java
```diff
@@ -0,0 +1,60 @@
...
+import com.ibm.streams.operator.model.Icons;
+import com.ibm.streams.operator.model.InputPortSet;
+import com.ibm.streams.operator.model.OutputPortSet;
+import com.ibm.streams.operator.model.PrimitiveOperator;
+import com.ibm.streamsx.topology.function.ToIntFunction;
+import com.ibm.streamsx.topology.internal.spljava.SPLMapping;
+
+@PrimitiveOperator
+@InputPortSet(cardinality = 1)
+@OutputPortSet(cardinality = -1)
```
ddebrunner: Isn't the default cardinality unlimited?

dlaboss: I do seem to recall some doc that said the default was unlimited, but in fact it is not.
removing the decl yields:

java.lang.IllegalStateException: Operator com.ibm.streamsx.topology.functional.java::FunctionSplit does not allow output ports.

java/src/com/ibm/streamsx/topology/TStream.java
dlaboss: related to that, shouldn't there also be a split(int n) that lacks a ToIntFunction, where, just like parallel, the behavior is: if the tuple supports Keyable<> that's used, otherwise it's round-robin.

also, shouldn't there be a parallel(..., ToIntFunction()), enabling arbitrary partitioning logic w/o the need to define new tuple types (and transforms) just to get different partitioning of the same logical tuple stream in different parallel regions?

wmarshall484: With respect to parallel(..., ToIntFunction()), are you thinking a use case might be something like:
```java
TStream<String> names = getTStreamOfNames();

// Partition based on whether name begins with "A"
TStream<String>  partitionWithA = names.parallel(2, new Function<String, Integer>(){
    @Override
    public Integer apply(String name){
        if(name.startsWith("A")){
            return 0;
        } 
        else{
            return 1;
        }
    }});

// Partition based on whether name begins with "B"
TStream<String>  partitionWithB = names.parallel(2, new Function<String, Integer>(){
    @Override
    public Integer apply(String name){
        if(name.startsWith("B")){
            return 0;
        } 
        else{
            return 1;
        }
    }});
...
```

Where in the above case `names`, the same logical tuple stream, is partitioned in two different ways without creating any kind of intermediate transformation.

java/src/com/ibm/streamsx/topology/internal/functional/ops/FunctionSplit.java
ddebrunner: Removing what exactly, I just meant the cardinality value?

dlaboss: ah, i removed the entire @OutputPortSet decl.

java/src/com/ibm/streamsx/topology/TStream.java
dlaboss: right, it makes that sort of thing possible.  I just noticed issue#114 which notes other inconveniences of the current scheme. We should probably warp over to that for further discussion on parallel().

Still interested in thoughts on whether split() should also offer a Keyable mode -- either a la issue#114 or implicit like today's parallel().  Maybe not.  Wonder if implements Keyable or issue#114 scheme really is a benefit over an explicit partitioning fn.  The splitter writer is also free to check (or know) if the type implements keyable and use that in their computation.  Mostly I was noting the dissimilarity in partitioning mechanisms between parallel() and split() and wondered if they should be more similar.

ddebrunner: Missed this conversation.

I don't think split() should offer a non-function version, it is different to parallel() and I think having split(n) and parallel(n) would add confusion as to which to use.
The only way a non-function (Keyable) version would work is to use the hash code of tuple's key, it seems unrealistic that the hash code % n will match any desired routing of tuples. For any normal hash code, won't ~50% of the tuples be discarded as its value will be less than 0?

PR #135, opened 2015-08-03
Big pull request to allow streams of parameterized types.

ddebrunner:
Can now support streams such as `TStream<List<String>>`


java/src/com/ibm/streamsx/topology/internal/logic/Constants.java
```diff
@@ -14,7 +18,19 @@
...
-        this.data = data;
+        List<T> copiedData;
+        
+        if (data.isEmpty())
+            copiedData = data;
+        if (data instanceof ArrayList || data instanceof LinkedList
+                || data instanceof Stack || data instanceof Vector)
+            copiedData = data;
+        else {
+            copiedData = new ArrayList<>(data.size());
```
wmarshall484: Curious as to why this kind of copying is necessary. Is there a case where it matters what kind of list it is?

ddebrunner: To ensure the list is serializable, for example a list returned from HashMap.values() is not serializable. I'll add a comment.

dlaboss: also looks like lines 23,24 are superfluous and should be removed.

PR #142, opened 2015-08-06
Adds a FunctionContext interface

ddebrunner:
Also adds endParallel() and fixes a union bug.

### DIFF
--- a/java/src/com/ibm/streamsx/topology/function/FunctionContext.java
+++ b/java/src/com/ibm/streamsx/topology/function/FunctionContext.java
@@ -72,7 +72,7 @@ public interface FunctionContext {
      * <P>
      * Any uncaught exception thrown by the {@code Runnable} passed
      * to the {@code ThreadFactory.newThread(Runnable)} will cause
-     * the processing element containing the operator to terminate.
+     * the processing element containing the function to terminate.
      * </P>
      * <P>
      * The ThreadFactory will be shutdown when the processing element is to be shutdown.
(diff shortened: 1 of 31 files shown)


java/src/com/ibm/streamsx/topology/function/FunctionContext.java
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Context for a function executing in a IBM Streams application.
+ */
+public interface FunctionContext {
+    
+    /**
+     * Get the processing element hosting a function 
+     * @return ProcessingElement hosting a function.
+     */
+    ProcessingElement getPE();
```
dlaboss: Is exposing com.ibm.streams.ProcessingElement a temporary thing?  
Should we create our own interface (and delagating impl) to keep the public api as decoupled as possible?

ddebrunner: Not sure we need to, the api elsewhere depends on the Java Operator API.

Do you think some of the ProcessingElement API's need to be hidden?

dlaboss: Hadn't really looked all that closely at it.  It was more out of a general observation.

Does the Java Operator API "leak through" elsewhere other than when explicitly using something in com.ibm.streamsx.topology.spl?  The public api seems overwhelmingly independent of the Java Operator API and doing something here seems like a small price to pay to keep it that way.

Unrelated to whether or not we return that interface, it seems like additional doc is needed somewhere to describe certain values under certain execution contexts - i.e., for StreamContext.STANDALONE / when ProcessingElement.isStandalone()==true, the behavior of getJobId(), getPEId(), getInstanceId(), getDomainId()

PR #162, opened 2015-08-19
Add Placeable interface without any implementation

ddebrunner:
Do not merge, for review only!

Just adding as a indication of the direction for supporting host placement.

I added a new interface Placeable which has the fuse and add resource tag methods. It allows continued fluent style as was raised as an issue in #9 , e.g.

`top.endlessSource(…).addResourceTag("ingest").filter(...).transform(...)….`

### DIFF
--- a/java/src/com/ibm/streamsx/topology/context/Placeable.java
+++ b/java/src/com/ibm/streamsx/topology/context/Placeable.java
@@ -7,6 +7,7 @@
 import java.util.Set;
 
 import com.ibm.streamsx.topology.TopologyElement;
+import com.ibm.streamsx.topology.builder.BOperatorInvocation;
 
 /**
  * Placement directives for a topology element when executing
@@ -21,6 +22,8 @@ import com.ibm.streamsx.topology.TopologyElement;
  */
 public interface Placeable<T extends Placeable<T>> extends TopologyElement {
     
+    boolean isPlaceable();
+    
     /**
      * Fuse this container with other topology elements so that
      * at runtime they all execute within the same operating system process.
@@ -31,7 +34,7 @@ public interface Placeable<T extends Placeable<T>> extends TopologyElement {
      * 
      * @return this
      */
-    public T fuse(Placeable<?> ... elements);
+    T fuse(Placeable<?> ... elements);
     
     /**
      * Add required resource tags for this topology element for distributed submission.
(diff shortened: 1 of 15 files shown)


ddebrunner:
Jenkins skip

java/src/com/ibm/streamsx/topology/context/Placeable.java
```diff
@@ -0,0 +1,51 @@
...
+     * 
+     * @param elements Elements to fuse with this container.
+     * 
+     * @return this
+     */
+    public T fuse(Placeable<?> ... elements);
+    
+    /**
+     * Add required resource tags for this topology element for distributed submission.
+     * This topology element and any it has been {@link #fuse(Placeable...) fused}
```
dlaboss: Looks good!  Seems like there should be some mention here, and above, of low-latency regions too.
And add a reference to this from the low-latency methods.

java/src/com/ibm/streamsx/topology/context/Placeable.java
```diff
@@ -0,0 +1,51 @@
...
+ * {@link StreamsContext.Type#ANALYTICS_SERVICE ANALYTICS_SERVICE}
+ * or {@link StreamsContext.Type#DISTRIBUTED_TESTER DISTRIBUTED_TESTER} context.
+ * <BR>
+ * For all other context types directives are ignored.
+ */
+public interface Placeable<T extends Placeable<T>> extends TopologyElement {
+    
+    /**
+     * Fuse this container with other topology elements so that
+     * at runtime they all execute within the same operating system process.
```
wmarshall484: Since running in the same process is the current default behavior, would this override the default? As in, if you have these TStreams:
A->B->C->D

and you call

B.fuse(C)

Would it be the case that A and D run on one process, while B and C run on another?

ddebrunner: The fusing constraints only provide guarantees about fusing, so in this case it is guaranteed that B and C will execute in the same process. Where A and D execute is not guaranteed.

java/src/com/ibm/streamsx/topology/context/Placeable.java
ddebrunner: Possibly, need to think about that.

ddebrunner:
IBM Streams describes these values as "Submission Time Values" though the JMX api is "SubmitParameter", so should the SubmissionParameter class be renamed?

ddebrunner:
Merging this as it brings some benefits to the handling of virtual markers. The fusing/host tag apis exist and are tested in terms of creating the correct graph elements, but the SPL implementation is still to be done.

PR #165, opened 2015-08-24
Add submission parameter support.

dlaboss:
- Add ability to specify that an SPL operator parameter has a submission
time value with or without a default value.
- Add abilty to express submission time parameter value in
StreamsContext.submit() for STANDALONE, DISTRIBUTED, ANALYTICS_SERVICE
- Add test cases to SPLOperatorTest

- N.B. these changes do not address providing functional logic with the
ability to access a submission time value, nor to specify something like
a submission time parameter for parallel region width.


dlaboss:
@ddebrunner, it's easy enough to add .parallel(new SubmissionParameter(...)) support -- which @hildrum asked for in issue#132.  Go for it?

java/src/com/ibm/streamsx/topology/spl/SubmissionParameter.java
```diff
@@ -0,0 +1,26 @@
+package com.ibm.streamsx.topology.spl;
+
+import com.ibm.streamsx.topology.builder.BSubmissionParameter;
+
+public class SubmissionParameter extends BSubmissionParameter {
```
ddebrunner: I think this needs to be a class that contains a BSubmissionParameter rather than extending it.

Also need class level comments, at least indicating how this is used.

test/java/src/com/ibm/streamsx/topology/test/spl/SPLOperatorsTest.java
```diff
@@ -119,6 +122,201 @@ public void testLiterals() throws Exception {
...
+//        assertEquals(ui32, tuple.getInt("ui32"));
+//        assertEquals(ui64, tuple.getLong("ui64"));
+
+        assertEquals("f32="+f32, f32, tuple.getFloat("f32"), 0.001);
+        assertEquals("f64="+f64, f64, tuple.getDouble("f64"), 0.001);
+    }
+
+    static void addParamDefault(String name, String splType, String defaultVal, Map<String,Object> params)
+    {
+        String spName = name + ".param";
```
ddebrunner: Why is the name of the parameter different in the map passed in at submission time?

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
```diff
@@ -319,7 +319,19 @@ static void paramClause(JSONObject graphConfig, JSONObject op,
     static void parameterValue(JSONObject param, StringBuilder sb) {
         Object value = param.get("value");
         Object type = param.get("type");
-        if (value instanceof String && !PARAM_TYPES_TOSTRING.contains(type)) {
+        if ("submissionParameter".equals(type)) {
+            JSONObject jo = (JSONObject) value;
+            String splType = (String) jo.get("splType");
+            String name = SPLGenerator.stringLiteral((String) jo.get("name"));
```
ddebrunner: Does a parameter name need escaping as a string literal?

ddebrunner:
I think this needs some evolution, SubmissionParameter may need to be an interface or at least not specific to SPL, especially if it's supported for @parallel. Also not sure if we need the BSubmissionParameter, just have SubmissionParameter implement JSONable. And should we have type specific versions SubmissionParameter, so that the user is not forced to convert their value to a String, e.g. IntSubmissionParameter, or maybe SubmissionParameter is parameterized, SubmissionParameter<T> ?

Isn't there also an issue with the actual name of the submission value is tied into how deeply the operator is nested, so it seems the parameter name could change based upon if it's in a parallel region or not. Maybe we can fix the parameter names by having them be composite parameters at the main composite level? But then it becomes more involved getting the value down to the operator invocation.

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
dlaboss: the value does need to be converted to a string literal so using the generator's method for that seemed logical/sensible.  though the value presumably won't/shouldn't ever have any chars that require the escapes that stringLiteral() currently handles, it doesn't do any harm.  i'll remove it if you think it's best w/o it.

test/java/src/com/ibm/streamsx/topology/test/spl/SPLOperatorsTest.java
dlaboss: just arbitrary in the test.  SPL getSubmissionTimeValue() doesn't have any requirement that they're the same.  i'll adj the test though.

java/src/com/ibm/streamsx/topology/spl/SubmissionParameter.java
dlaboss: The code started out that way :-)   The issue was that topology.builder.BOperatorInvocation.setParameter() would then end up with a dependency on SubmissionParameter, which seemed best avoided since topology.builder seemed to have virtually no other dependencies on other streamsx.topology entities, or that callers of BOperatorInvocation(...,params) would need to first massage the params to cvt to a BSubmissionParameter, which seemed unnecessarily cumbersome/error prone.  I could convert BSubmissionParameter to an interface though I don't think that fundamentally changes things.  But I'll make the change to that you suggested if that's your preference.  Let me know.

dlaboss:
I should have read all the comments before responding to some :-)   
- yup, using SubmissionParameter in parallel() implicitly included migrating SubmissionParameter to streamsx.topology package.
- re: BSubmissionParameter elimination via JSONable SubmissionParameter: sounds good
- re: "default value" as a String, perhaps I was overly influenced by the fact that SPL requires this.  A parameterized SubmissionParameter idea seems promising and since we're bound  to java8 presumably we'd allow unsigned int values for SPL unsigned types (we need something to generate the cast in the generated SPL code).  How would it expose expressing that it was for a ustring type vs rstring?  What about complex SPL types?
- re: actual / fully qualified param names affected by op containment hierarchy, yeah, that characteristic of SPL seems unfortunate and requires more thought if we really want/need to try to isolate the java api user from this... without support from the SPL runtime/compiler.  e.g., it also providing a simple "developer must define unique submission param names" model... where then also multiple operators could reference any submission param regardless of the operator's SPL hierarchy (a la java properties).  Maybe just leave this particular issue/concern until a time when Streams does/can provide support for that?

hildrum:
I like the idea of a parameterized SubmissionParameter, since as an SPL user, I've found it irritating that getSubmissionTimeValue expects strings which have to be converted to other types.

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
ddebrunner: Leave as-is, makes sense with  your explanation.

ddebrunner:
I think like SPLStream extends TStream, a SPL specific version of SubmissionParameter might exist, that includes an SPL type.

ddebrunner:
FYI, the api still supports Java 7, so it is not bound to Java 8.

dlaboss:
jenkins please test

dlaboss:
Added support for parallel() with submission time value.  See the commit's comment for details.
@ddebrunner, assuming you like the way this ended up, should we make the "old" TStream.parallel() methods deprecated in order to simplify the API?

ddebrunner:
I'm not sure that removing the existing parallel methods simplifies the api, now the user must do:

`s.parallel(new Parameter<Integer>(3))`
instead of
`s.parallel(3)`

I'm also struggling a little with the name 'Parameter', as it's not a parameter, it's just a value wrapper, and if Parameter needs to have knowledge of SubmissionParameter, should instead SubmissionParameter extend Parameter?

Also, will we need to have every method that takes a value, now overloaded to take Parameter?

Just trying to see how this will work as we expand the use of submission time parameters.

ddebrunner:
Once the code has been finalized can you update the JSON graph format description :

https://github.com/IBMStreams/streamsx.topology/wiki/JSON-Format-for-Graphs

I think it's just the parameter object section.

dlaboss:
- re: "simplifies", I should have said "reduces the API's footprint / surface area".  Smaller / fewer options => typically easier to comprehend => simpler :-)
- re: "Parameter", perhaps "Value" would be better name?  I like that / will change it.  Yeah, SubmissionParameter being a kind of value seems like a good idea.
- re: usage/overloads, do you see any other option?  At least with Value, anything new supporting submission time params would presumably only define a Value type'd variant.  And anything existing that's updated to support it has the opportunity to deprecate/rm the original - in service of "reducing the footprint".

ddebrunner:
IBM Streams describes these values as "Submission Time Values" though the JMX api is "SubmitParameter", so should the SubmissionParameter class be renamed?

dlaboss:
jenkins please test

ddebrunner:
-1 on this being merged at this point.

Need to better understand the path we are heading down.

java/src/com/ibm/streamsx/topology/SubmissionParameter.java
```diff
@@ -0,0 +1,185 @@
...
+     * @param name submission parameter name
+     * @param valueClass class object for {@code T}
+     * @throws IllegalArgumentException if {@code name} is null or empty
+     */
+    public SubmissionParameter(String name, Class<T> valueClass) {
+        if (name == null || name.trim().isEmpty())
+            throw new IllegalArgumentException("name");
+        this.name = name;
+        this.valueClass = valueClass;
+        this.defaultValue = null;
```
wmarshall484: Here, defaultValue is allowed to be null. In the other constructor, on line 89, a null defaultValue throws an exception. Is there a use case where a user would want a SubmissionParameter with a null value?

java/src/com/ibm/streamsx/topology/SubmissionParameter.java
```diff
@@ -0,0 +1,184 @@
...
```
wmarshall484: Here, defaultValue is allowed to be null. In the other constructor, on line 89, a null defaultValue throws an exception. I feel like invoking 

```java
String s = null;
SubmissionParameter<String> param = new SubmissionParameter<String>("someName", s);
```
would be the equivalent of invoking

```java
SubmissionParameter<String> param = new SubmissionParameter<String>("someName", String.class);
```

Is there a reason that the latter should set a null defaultValue, while the former doesn't?

ddebrunner: I don't think the one takes String.class has a default value of null, instead it has no default value., so a value must be supplied on submission. (yes defaultValue is set to null, but that means there is no default value).

ddebrunner:
Just struck me that Value is very much like Supplier, I wonder if there should be a more functional approach to this.   `parallel(Supplier<Integer>)`

`<T> Supplier<T> getSubmissionValue(String name, T defaultValue);`

dlaboss:
Trying to understand all the ideas surrounding parallel(Supplier<Integer>) and Supplier<T> getSubmissionParameter().  Seems reasonable but want to be sure...

- Would getSubmissionParameter(...) be a method on Topology?  E.g.
  - stream.filter(...).parallel(top.getSubmissionParameter<Integer>("p1",...))
- Will parallel(Supplier) accept any arbitrary Supplier (e.g., a user defined Supplier)... with the restriction that for anything other than a getSubmissionParameter() Supplier, Supplier.get() will be called at compile time and must return a non-null value?
- Presumably, SubmissionParameter becomes internal.
- Presumably, an SPL op param spec would now accept a Supplier.  If it's a non-submission-param Supplier, a user defined Supplier, it has the same requirements/semantics for use in the parallel(Supplier) case above.
- What happens if a getSubmissionParameter(...) returned Supplier is passed into other places in the API that take a Supplier?  e.g., would this be expected to work - i.e., Supplier.get() is called at runtime and it returns the runtime value?  Off hand, not sure of the implications of that :-)  e.g. can we make that work with today's underpinnings - SPL/runtime/operator-api? e.g.
  - top.source(top.getSubmissionParameter<Foo>("p1",...))

ddebrunner:
Yeah I think getSubmissionParameter would be a method on topology.

Supplier.get would be called at StreamsContext.submit time.

We could probably make `top.sourceXXX(top.getSubmissionParameter("p1",...))` work, though may not be needed at first.

What's interesting with lambda functions, you could pass:

parallel(()->3)

almost as simple as parallel(3).

ddebrunner:
The parameter name still needs to be resolved, as the api is written, I would expect multiple uses of a named submission value  to refer to the same value, but the SPL generation will not result in that.

ddebrunner:
> Supplier.get would be called at StreamsContext.submit time.

Though the more intuitive approach is that is is called when the application executes, as that's when other functions are called.

dlaboss:
The commit has all of the refactoring for Supplier<T> Topology.getSubmissionParameter(...).

The earlier public SubmissionParameter class had support for SPL ustring and unsigned ints though it was special-cased for submission parameters.  General support for SPL ustring and unsigned ints is now included (also addresses issue#89).  See new cist.spl.UString and Unsigned.  SPLOperatorsTest parameter tests have been extended to include those types, as literals as well as submission params.

Open:
- renaming getSubmissionParameter as @ddebrunner asked above?  Seems sensible, just need to pick.
- parallel(Supplier<T>, ...) Supplier.get() is called when parallel() is called, not as part of submit()

ddebrunner:
Isn't there still the open issue of what is the actual name of the parameter at bundle submission time?

dlaboss:
Uh, yep.  Is that something we're going to tackle now - pre-merge? (if so, how?)
Or minimally add doc about (a) choosing unique names for logically distinct properties (then being able to specify just the leaf name at submission time), and (b) that at this time multiple uses/references of the same logical param (leaf name) can result in independent parameters if the params are used in different "composites" -- need to describe that in terms of parallel regions.

ddebrunner:
Is it just different composites, or different operator invocations as well?

dlaboss:
ok, here's the scoop w/SPL

SPL: each composite definition (not instance) introduces a new STV namespace having the composite's full name.   
- ``getSTV(name)`` always interprets ``name`` as a leafname in the calling composite's STV namespace, hence the STV param's full name is: ``<full-composite-def-name>.<leafname>``.  
- Within a composite, ``getSTV(some-name)`` can be called as many times as desired; it's the same param and the compiler enforces that all calls in the composite have the same default-value specification.  
- Sharing of a single STV among composites is achieved by the use of composite parameters.

FWIW: streamtool, and I think the Application Service REST API, not sure about the Console, allow specifying just the leafname as long as it's unique in the app.  Studio presents the user with a table of all of the app's STVs.

Proposal will follow...

dlaboss:
API proposal (``SP`` is shorthand for ``SubmissionParameter``):
- ``Topology.getSP("some-name")`` defines a new SP name that's global to the topology;  throws if the name has already been defined.  Consider adding?:
  - ``Topology.getSPNames() // returns a List<String> of names from successful getSP() calls``
- Until there's a proven need, there's no way for an API generated non-main composite to define its own SP namespace / params.
- Multiple uses of a SP are achieved by reusing the ``Supplier`` returned by ``getSP()``.  Consider adding?:
  - ``Topology.lookupSP(String name) // returns the Supplier<?> for the name; null if name not yet defined.``

API design notes (roughly):
- ``top.getSP("myName", 10)`` induces generating in the main composite:

    ```expression<the-type> $jaa_stv_myName : (int32) getSubmissionTimeValue("myName", "10");```
- uses of the returned ``Supplier`` in API generated non-main composites induces generating in the composite:

    ```expression<the-type> $jaa_stv_myName : (int32) 10;  // param is optional```
- uses of the returned ``Supplier`` in any composite are replaced with ``$jaa_stv_myName``.

ddebrunner:
I don't think we need `Topology.lookupSP(String name)`, it's the same function as `Topology.getSP(String name)`.

hildrum:
Do you have an example of how this would be used to send a submission time parameter to a java primitive operator?

dlaboss:
You'll just specify a ``Topology.getSubmissionParameter(...)`` returned ``Supplier`` for a parameter's value. e.g.,
```
Supplier<Integer> submitParamFoo = topology.getSubmissionParameter("foo", 10);
...
Map<String,Object> params = new HashMap<>();
// params.put("aIntParam", 10);
params.put("aIntParam", submitParamFoo);
... = JavaPrimitive.invokeJavaPrimitive(..., params);
```

hildrum:
FWIW, I am happy with this API.

java/src/com/ibm/streamsx/topology/Topology.java
```diff
@@ -698,4 +699,68 @@ public GraphBuilder builder() {
 
         return names;
     }
+
+    /**
+     * Create a submission parameter without a default value.
+     * <p>
+     * A submission parameter is a handle for a {@code T} whose actual value
+     * is not defined until topology execution time.
```
ddebrunner: submission time, not "execution time"

java/src/com/ibm/streamsx/topology/Topology.java
```diff
@@ -698,4 +699,68 @@ public GraphBuilder builder() {
...
+     * <p>
+     * The returned Supplier's {@code get()} returns null.
+     * <p>
+     * Submission parameters may be used for values in various
+     * cases such as {@code TStream.parallel()} width specifications.
+     * e.g.,
+     * <pre>{@code
+     * TStream<Foo> s = ...
+     * s.parallel(top.createSubmissionParameter(..., 5) ... // default width of 5
+     * s.parallel(top.createSubmissionParameter(..., Integer.class) ...
```
ddebrunner: Somewhat confusing having two parallel calls against the same stream, and maybe too many '...', leaving too much guesswork for the reader.

Why not just have a complete example?

java/src/com/ibm/streamsx/topology/Topology.java
```diff
@@ -698,4 +699,68 @@ public GraphBuilder builder() {
...
+     * }</pre>
+     * <p>
+     * Submission parameters may be used in Java Primitive Operator and
+     * SPL Operator parameter specifications.
+     * <p>
+     * The name is used to supply an actual value at topology execution time
```
ddebrunner: submission

ddebrunner:
+1 to merging this to make progress (once it passes the checks). I don't think we need to introduce a new type system for spl types (UString and the Unsigned types), given the spl portion is built on top of IBM Streams's Java Operator API which already has a full SPL type system.

I was maybe expecting a specialization of submission parameters for SPL values that maybe took a Type.MetaType parameter to indicate specific SPL types.

Addition comments can also be handled in a follow up.

java/src/com/ibm/streamsx/topology/internal/core/StreamImpl.java
```diff
@@ -362,13 +363,25 @@ public void publish(String topic) {
         this.connectTo(op, false, null);
     }
     
-    public TStream<T> parallel(int width, Routing routing) {
+    public TStream<T> parallel(final int width, Routing routing) {
+        @SuppressWarnings("serial")
+        Supplier<Integer> supplier = new Supplier<Integer>() {
```
ddebrunner: Isn't there an issue here that Supplier captures the TStream instance, and so is not actually serializable?

java/src/com/ibm/streamsx/topology/internal/core/StreamImpl.java
```diff
@@ -362,13 +363,25 @@ public void publish(String topic) {
...
-        if (width <= 0) {
-            throw new IllegalStateException(
-                    "The parallel width must be greater"
-                            + " than or equal to 1.");
-        }
+        if (width == null)
+            throw new IllegalArgumentException("width");
+        Integer widthVal = width.get() != null
+                ? width.get()
+                : ((SubmissionParameter<Integer>)width).getDefaultValue();
```
ddebrunner: This will throw an ClassCastException is the supplier returns null and is not a submission value.

dlaboss:
post-merge comments will be addressed in a separate branch/pull

dlaboss:
All of the review comments are addressed in pull#182

PR #180, opened 2015-09-02
Added cat's cradle functinality and tests.

wmarshall484:
The hashAdder object is now moved to before the $endParallel$ operator when there are two adjacent parallel regions. E.g., whenever you call endParallel().parallel() with partitioned routing (either hash or key):

```java
stream.parallel(5, TStream.routing.HASH_PARTITIONED).filter(...).endParallel().parallel(TStream.routing.HASH_PARTITIONED).
```
There are several edge cases, each of which is covered with a test added to ParallelTest.


java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -135,6 +137,51 @@
...
+     * @param op
+     * @param name
+     */
+    static JSONObject copyOperatorNewName(JSONObject op, String name){
+	try{
+	    op = OrderedJSONObject.parse(op.serialize());
+	}
+	catch(Exception e){
+	    e.printStackTrace();
+	    System.exit(-1);
```
ddebrunner: Don't call System.exit(), wrap the Exception in a RuntimeException and throw it.

java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -135,6 +137,51 @@
...
+        for(JSONObject output : outputs){
+            output.put("name", name + "_OUT" + output.get("index").toString());
+            JSONArray conns = (JSONArray) output.get("connections");
+            conns.clear();
+        }
+        return op_new;
+    }
+
+    static void removeOperator(JSONObject op, JSONObject graph){
+	List<JSONObject> opList = new ArrayList<>();
```
ddebrunner: Collections has a singletonList method.

java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -135,6 +137,51 @@
...
+	catch(Exception e){
+	    e.printStackTrace();
+	    System.exit(-1);
+	}
+        JSONObject op_new=null;
+        try {
+            op_new = JSONObject.parse(op.serialize());
+        } catch (IOException e) {
+            e.printStackTrace();
+            System.exit(-1);
```
ddebrunner: DOn't call System.exit

java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -135,6 +137,51 @@
...
+    
+    /**
+     * Copies operator, giving it a new name. Also renames input/output ports,
+     * and clears the input/output connections array.
+     * @param op
+     * @param name
+     */
+    static JSONObject copyOperatorNewName(JSONObject op, String name){
+	try{
+	    op = OrderedJSONObject.parse(op.serialize());
```
ddebrunner: Can this be explained in comments, why deserializing twice is needed?

test/java/src/com/ibm/streamsx/topology/test/api/ParallelTest.java
```diff
@@ -43,6 +43,85 @@
...
+	checkUdpSupported();
+	Topology topology = new Topology("testCat");
+	TStream<String> out = topology.strings("hello").parallel(5, TStream.Routing.HASH_PARTITIONED).filter(nopStringFilter()).endParallel()
+	    .parallel(5, TStream.Routing.HASH_PARTITIONED).filter(nopStringFilter()).endParallel();
+
+	Tester tester = topology.getTester();
+	Condition<Long> expectedCount = tester.tupleCount(out, 1);
+        Condition<List<String>> hello = tester.stringContents(out, "hello");
+
+	StreamsContextFactory
```
ddebrunner: Can we change these to use the standard approach for writing tests, so they are submitted to the context provided by the TestTopology super class.

test/java/src/com/ibm/streamsx/topology/test/api/ParallelTest.java
```diff
@@ -612,6 +691,16 @@ public void testParallelIsolate() throws Exception {
     }
     
     @SuppressWarnings("serial")
+    public static Predicate<String> nopStringFilter(){
```
ddebrunner: FYI, there is already a parameterized AllowAll filter.

test/java/src/com/ibm/streamsx/topology/test/api/ParallelTest.java
```diff
@@ -43,6 +43,85 @@
 import com.ibm.streamsx.topology.tuple.BeaconTuple;
 
 public class ParallelTest extends TestTopology {
+
+    @Test
+    public void testCatsCradle() throws Exception {
+	checkUdpSupported();
+	Topology topology = new Topology("testCat");
+	TStream<String> out = topology.strings("hello").parallel(5, TStream.Routing.HASH_PARTITIONED).filter(nopStringFilter()).endParallel()
```
ddebrunner: Is there a simple way to have tests that actually test the parallelism requested was implemented?

By just passing a single tuple through it seems they would pass if the implementation actually didn't use parallelism?

Maybe the functions being parallelized could actually test the max channel number?

dlaboss:
Perhaps I'm just being dense but I guess I don't understand why adjacent/abutting parallel regions affects the API code in a "cats cradle" sense.  i.e., we're just generating the SPL with parallel annotations on adjacent ops.  The SPL compiler and the runtime are responsible for all of the gore... replication/splitters/cats-cradle physical op interconnects.    Would it be more appropriate to recast the description of all of this as simply adjacent/abutting parallel regions and dispense with the "cat's cradle" terminology in our code / tests?  If that wouldn't be appropriate, maybe there needs to be some sufficiently detailed doc / example in the code to make all of this clearer?

ddebrunner:
@dlaboss I'm probably being denser by I don't see any changes to the API so how is this affecting the API code?

dlaboss:
@ddebrunner sorry, didn't mean to imply changes to the API.  Was just saying there's a lot of impl changes (``GraphUtilities``, ``Preprocessor``) and tests that are described in service of "cats cradle" and it's not obvious to me how / why "cats cradle" affects the API impl.  Hence is this all more simply in service of "abutting parallel regions"?  Or what really is the "cats cradle" functionality that has to be implemented?

wmarshall484:
Just to make sure I understand, you're saying that cat's cradle functionality is a feature that's handled by the SPL runtime, not our API/impl. Therefore, we should only comment how these changes enable adjacent/abutting parallel regions, and *not* that it enables cat's cradle at the SPL level.

dlaboss:
Right.  Presumably, if we can generate SPL with two directly connected ``@parallel`` (composite) ops (from ``... endParallel().parallel(...)...``) then we've done all we need to do and there's nothing cats-cradle-ish about that work -- or we have different understandings of what that means :-)

PR #189, opened 2015-09-04
Add mqtt

dlaboss:
- add messaging.mqtt package, mqtt.ConsumerConnector/ProducerConnector (following our kafka pattern), tests, sample.  Support submission parameter based connector and topic configs.
- add Logic.Value<T>  -- a Supplier<T> (moved pvt impl from StreamImpl)
- fix/tweak some sample doc

Internal:
- fix SubmissionTimeValue to handle "." in parameter names
- enhance BOperatorInstance.setParamter() to handle Supplier<T> -- allow
general Supplier<T> to be an op param value.
- relocate tagOpAsJavaPrimitive to SPL (one copy used by any of our code that needs it)


dlaboss:
fixes issue#158

PR #214, opened 2015-09-16
Support Submission Parameter Supplier use in functional logic.

dlaboss:
General approach / changes:
- fixes issue#199
- add functional.ops.SubmissionParameterManager
  - A singleton that provides the collection of STV values at runtime.
  - its getValue(spName, metaType) is called by SP Supplier.get()
  - STANDALONE & DISTRIBUTED: initialized from operator context info
(values are retrieved from the op context's new "submissionParams"
parameter)
  - EMBEDDED: initialized via graph's submission parameter params and
the ContextProperties.SUBMISSION_PARAMS info
- add "rstring[] submissionParamNames" and "rstring[] submissionParamValues" functional operator parameters.  When
SPs are present the param is set to convey all SPs. e.g.,
  - submissionParamNames: "__jaa_stv_threshold", "__jaa_stv_width"
  - submissionParamValues: (rstring) $__jaa_stv_threshold,  (rstring) $__jaa_stv_width
- ensure composite defs/instances make all submission params available
as required.
- the graph's json is unchanged as a result of these changes.  With the exception of logically minor changes to `internal.core.SubmissionParameter`, essentially all of the changes are in `internal.functional.ops` and  `generator.spl`

- enhance ParallelTest to verify SP in functional logic within a region
works.
- add FunctionalSubmissionParamsTest


ddebrunner:
Interesting approach.

I don't see any api documentation on how these are used, did I miss it?

The use of a single static map in SubmissionParameterManager is an issue, it assumes there is only a single topology executing at a time, which may not be the case. Also possible issues with memory leaks, if the values in the map are never removed.

dlaboss:
Duh forgot the doc.
>1 topology in a single class loader context? Why would we do/want that?  But shouldn't be a big deal to change to a per-topology manager.  Though I'm inclined to defer that until we (underlying runtimes) support such a config.  Do you agree?

ddebrunner:
We do support such a configuration, with embedded multiple topologies can be executing.

Typically, static (modifable) variables are not a good idea in Java as it limits the use in embedded cases, which should be the goal for all Java libraries.

dlaboss:
created issue#218 to track the concurrent executing embedded topologies issue.

PR #220, opened 2015-09-23
Issue 181 coloc breaks low latency, and coloc breaks isolate, and isolate breaks low latency

### LINKED ISSUE #181: Colocating a stream with a low latency section breaks the low latency guarantee
In distributed the low latency operators end up in different PEs.

dlaboss:
fixes #181 - colocate breaks low latency, and colocate breaks isolate, and isolate breaks low latency

- disallow colocate within a low latency region (will followup with a
new issue to support it)
- disallow colocate() violating isolate()
- disallow isolate in a low latency region
- make PlacementInfo.colocate() threadsafe
- make PlacementInfo.colocate() check constraints on all args
- add useful exception message text
- add test for isolate() in a low latency region
- add tests for colocate with low latency region, colocate with isolate
- add tests for some "not placeable" colocate usage (will followup with
issue to track improving usability)
- add test for nested low latency regions

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -31,6 +31,14 @@
     </ant>
   </target>
 
+  <target name="test.everything" >
+    <ant dir="test/java">
+       <target name="unittest.everything"/>
+       <target name="junit.report"/>
+       <target name="coverage.report"/>
+    </ant>
+  </target>
+
   <target name="test.quick" >
     <ant dir="test/java">
        <target name="unittest.quick"/>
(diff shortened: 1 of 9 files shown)


ddebrunner:
> make PlacementInfo.colocate() threadsafe

What's this for? I'm not sure we guarantee than building the graph is thread safe.

dlaboss:
seemed like the right thing to do since addResourceTags() and getResourceTags() were already synchronized... and colocate() directly reads/writes the same underlying resourceTags.

PR #224, opened 2015-09-25
@ignore FunctionalSubmissionParamsTest unit test that's suddenly failing

dlaboss:
failing only on jenkins streams.topology build stream.


ddebrunner:
Can you add a bug issue to track the failure.

Issue #211, opened 2015-09-16
cryptic IAE with TWindow.aggregate() and sub-1msec time


dlaboss:
Running in EMBEDDED_TESTER...

```
        s.last(2).aggregate(functionListFn(someInt, 100), 100, TimeUnit.MICROSECONDS);
        // the following is OK
        // s.last(2).aggregate(functionListFn(someInt, 100), 1, TimeUnit.MILLISECONDS);
        ...
        complete(tester, expectedCount, 15, TimeUnit.SECONDS);
```

yields

```
    <error message="java.lang.IllegalArgumentException" type="java.util.concurrent.ExecutionException">java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException
    at java.util.concurrent.FutureTask.report(FutureTask.java:134)
    at java.util.concurrent.FutureTask.get(FutureTask.java:214)
    at com.ibm.streamsx.topology.internal.tester.TupleCollection.complete(TupleCollection.java:489)
    at com.ibm.streamsx.topology.test.TestTopology.complete(TestTopology.java:146)
    at com.ibm.streamsx.topology.test.api.FunctionalSubmissionParamsTest.MiscTest(FunctionalSubmissionParamsTest.java:226)
Caused by: java.lang.IllegalArgumentException
    at java.util.concurrent.ScheduledThreadPoolExecutor.scheduleAtFixedRate(ScheduledThreadPoolExecutor.java:574)
    at com.ibm.streams.operator.internal.window.WindowHandler.scheduleRepeating(WindowHandler.java:264)
    at com.ibm.streams.operator.internal.window.WindowPartition$PeriodicTimerPolicy.activate(WindowPartition.java:263)
    at com.ibm.streams.operator.internal.window.sliding.SlidingCountEvictTimeTrigger.activate(SlidingCountEvictTimeTrigger.java:54)
    at com.ibm.streams.operator.internal.window.WindowHandler.activatePartitions(WindowHandler.java:392)
    at com.ibm.streams.operator.internal.window.WindowHandler$3.doWork(WindowHandler.java:385)
    at com.ibm.streams.operator.internal.window.WindowHandler$BackgroundWindowTask.run(WindowHandler.java:341)
    at com.ibm.streams.operator.internal.window.WindowHandler$2.handleNotification(WindowHandler.java:288)
    at javax.management.NotificationBroadcasterSupport.handleNotification(NotificationBroadcasterSupport.java:286)
[...]

PR #230, opened 2015-09-28
Carry the time unit for window & aggregations all the way to the JSON.

ddebrunner:


test/java/src/com/ibm/streamsx/topology/test/api/WindowTest.java
```diff
@@ -305,6 +321,32 @@ public void testPeriodicAggregateLastSeconds() throws Exception {
...
     
+    /**
+     * Test a periodic aggregation with microsecond aggregation.
+     * Basically a test that the application runs, hard to test
+     * for specific results for such a short window.
+     */
+    @Test
+    public void testPeriodicAggregateLastMicroseconds() throws Exception {
+        
+        assumeTrue(!isEmbedded());
```
dlaboss: I think it would help to add a comment why this is here - e.g., issue #211?

PR #237, opened 2015-09-29
Fixed adjacent parallel test. Added test.

### LINKED ISSUE #149: Support "cats-cradle" parallelism.
Allow topologies that include sequences similar to this:

`s.endParallel().parallel(4)`

including the various variants of parallel(...).

One issue is that a HashAdder added by the second parallel must be actually put into the preceding region, rather than after it, otherwise it's not a cats-cradle, but instead single threaded through the HashAdder.

wmarshall484:
Fixes #149


wmarshall484:
jenkins test this

wmarshall484:
jenkins test this

wmarshall484:
jenkins test this

PR #253, opened 2015-10-13
Remove use of common-build.xml from scala sample build

ddebrunner:


ddebrunner:
jenkins test this please

PR #265, opened 2016-01-27
Python spl

sdickes:
Dan, Will, 

Here is my small change for the Isolate virtual operator.

Steve

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
@@ -18,7 +18,10 @@ class SPLGraph(object):
     def addOperator(self, kind, function=None, name=None):
         if name is None:
             name = self.name + "_OP"+str(len(self.operators))
-        op = SPLInvocation(len(self.operators), kind, function, name, {}, self)
+        if (kind == "$Isolate"):
+            op = SPLMarker(len(self.operators), kind, name, {}, self)
+        else:
+            op = SPLInvocation(len(self.operators), kind, function, name, {}, self)
         self.operators.append(op)
         # TODO
         if not function is None:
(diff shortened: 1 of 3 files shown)


ddebrunner:

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -18,7 +18,10 @@ def __init__(self, name=None):
     def addOperator(self, kind, function=None, name=None):
         if name is None:
             name = self.name + "_OP"+str(len(self.operators))
-        op = SPLInvocation(len(self.operators), kind, function, name, {}, self)
+        if (kind == "$Isolate"):
```
ddebrunner: Rather than hardcoding Isloate, should it be starts with '$' ?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -188,3 +191,44 @@ def getSPLOutputPort(self):
...
+    def generateSPLOperator(self):
+        _op = {}
+        _op["name"] = self.name
+
+        _op["kind"] = self.kind
+        _op["partitioned"] = False
+
+        _op["marker"] = True
+        _op["model"] = "spl"
+        _op["language"] = "java"
```
ddebrunner: Probably should not be java here.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
sdickes: that makes sense. I will make that change once I determine my GitHub usage issues

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
sdickes: "python"?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -188,3 +191,44 @@ def getSPLOutputPort(self):
         _oport["name"] = self.name
         _oport["connections"] = [port.name for port in self.inputPorts]
         return _oport
+
+class SPLMarker(object):
```
ddebrunner: Technically this is not an SPLMarker, since it's not an SPL artifact. maybe just Marker

sdickes: okay, sounds good

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
ddebrunner: It's a marker "operator" that doesn't really have a language, the idea is that an operator in the graph has:

* model - the programming model of the operator
* kind - the kind of the operator within the programming model
* language - Implementation language

E.g. at some point a filter in Java would be:
 {topology, filter, java}
or Python
{topology, filter, python}

and later once it was translated into SPL the first might be:
{spl, JavaFunctionalFilter, spl}   // Can't rememeber the actual name of the operator

That's the goal, I think the graph currently dives to quickly into SPL, rather than being a later transformation step.

ddebrunner: Check with @wmarshall484 if the language is used for a marker.

ddebrunner:
Merging with the expectation there will be a follow up merge request to address the comments.

PR #269, opened 2016-02-12
Adding lowlatency/endlowlatency

sdickes:
[WIP] lowlatency/endlowlatency virtual operators.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
@@ -69,6 +69,8 @@ class SPLInvocation(object):
         self.inputPorts = []
         self.outputPorts = []
 
+        self.regions = []
+
     def addOutputPort(self, name=None, inputPort=None, schema= CommonSchema.Python):
         if name is None:
             name = self.name + "_OUT"+str(len(self.outputPorts))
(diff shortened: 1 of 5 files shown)


ddebrunner:

ddebrunner:
What's the purpose of regions in the python code?

sdickes:
Good question and I do not know if I have the full answer but when looking through the the java code for lowlatency and parallel operators and at the resultant json ouptut I see a region that gets added that has the name of the virtual operator. 
For example, I see this at line 104 in "java/src/com/ibm/streamsx/topology/builder/GraphBuilder.java"

BOutput lowLatencyOutput = addPassThroughMarker(parent, BVirtualMarker.LOW_LATENCY, true);

and the last argument controls the creation of the region, from what I can see. 

line 188 has this 

        if (createRegion) {
            final String regionName = op.op().getName();
            regionMarkers.put(regionName, virtualMarker.kind());
            op.addRegion(regionName);
        }

ddebrunner:
@wmarshall484 

Any thoughts on if regions are needed at this point?

I would have expected they are added during the processing of the markers, in the existing java code.

wmarshall484:
Regions are not needed in the python code. They are used by the SPL generator to hold information about operations like PE placement. With respect to generating correct parallel/lowLatency/isolate regions from python code, it's only important that the kinds are correctly named, e.g.:
``` Java
public enum BVirtualMarker {
    UNION("$Union$"),
    PARALLEL("$Parallel$"),
    END_PARALLEL("$EndParallel$"),
    LOW_LATENCY("$LowLatency$"),
    END_LOW_LATENCY("$EndLowLatency$"),
    ISOLATE("$Isolate$");
...
}
```

PR #272, opened 2016-02-24
Drhvm.parallel

drhvm:
Parallel - Unparallel using RoundRobin tuple distribution. Also contains lock fix for PythonfunctionSource when end of tuples is reached.  Changed OperatorGenerator.java to handle parallel width passed as a Long vs Int.  Also contains first pass on union but does not seem to work yet due to code gen problem?

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -34,7 +34,8 @@ class Topology(object):
         oport = op.addOutputPort(schema=schema)
         topicParam = {"topic": [topic]}
         op.setParameters(topicParam)
-        return Stream(self, oport)
+        return Stream(self, oport)    
+    
 
 class Stream(object):
     """
@@ -159,4 +160,45 @@ class Stream(object):
         op.addInputPort(outputPort=self.oport)
         oport = op.addOutputPort()
         return Stream(self.topology, oport)
+    
+    def parallel(self, width, func=None):
+        """
+        Marks operator as parallel output with width
+        :param: width
+        :returns: Stream
+        """
+        iop = self.isolate()
+               
+        op2 = self.topology.graph.addOperator("$Parallel$",func)
+        # addRegions()
+        op2.addInputPort(outputPort=iop.getOport())
+        oport = op2.addOutputPort(width)
+        return Stream(self.topology, oport)
 
+    def endParallel(self):
+        """
+        Marks end of operators as parallel output with width
+        :returns: Stream
+        """
+        op = self.topology.graph.addOperator("$EndParallel$")
+        # pop or addRegions()
+        op.addInputPort(outputPort=self.oport)
+        oport = op.addOutputPort()
+        endP = Stream(self.topology, oport)
+        return endP.isolate()
+
+    def union(self, streamList, schema=None):
+        """
+        The Union operator merges the streams that are connected to 
+        multiple input ports into a single stream.
+        :returns Stream
+        """        
+        op = self.topology.graph.addOperator("$Union$")
+        op.addInputPort(outputPort=self.getOport())
+        for stream in streamList:
+            op.addInputPort(outputPort=stream.getOport())
+        oport = op.addOutputPort()
+        return Stream(self.topology, oport)
+    
+    def getOport(self):
+        return self.oport
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -174,13 +174,20 @@ class Stream(object):
         oport = op2.addOutputPort(width)
         return Stream(self.topology, oport)
 
-    def endparallel(self):
+    def end_parallel(self):
         """
         Marks end of operators as parallel output with width
         :returns: Stream
         """
+        lastOp = self.topology.graph.getLastOperator()
+        outport = self.oport
+        if (isinstance(lastOp, graph.Marker)):
+            if (lastOp.kind == "$Union$"):
+                pto = self.topology.graph.addPassThruOperator()
+                pto.addInputPort(outputPort=self.oport)
+                outport = pto.addOutputPort()
         op = self.topology.graph.addOperator("$EndParallel$")
-        op.addInputPort(outputPort=self.oport)
+        op.addInputPort(outputPort=outport)
         oport = op.addOutputPort()
         endP = Stream(self.topology, oport)
         return endP.isolate()
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -174,7 +174,7 @@ class Stream(object):
         oport = op2.addOutputPort(width)
         return Stream(self.topology, oport)
 
-    def endParallel(self):
+    def endparallel(self):
         """
         Marks end of operators as parallel output with width
         :returns: Stream
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -161,7 +161,7 @@ class Stream(object):
         oport = op.addOutputPort()
         return Stream(self.topology, oport)
     
-    def parallel(self, width, func=None):
+    def parallel(self, width):
         """
         Marks operator as parallel output with width
         :param: width
@@ -169,8 +169,7 @@ class Stream(object):
         """
         iop = self.isolate()
                
-        op2 = self.topology.graph.addOperator("$Parallel$",func)
-        # addRegions()
+        op2 = self.topology.graph.addOperator("$Parallel$")
         op2.addInputPort(outputPort=iop.getOport())
         oport = op2.addOutputPort(width)
         return Stream(self.topology, oport)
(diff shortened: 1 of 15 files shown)


ddebrunner:

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
```diff
@@ -93,8 +93,12 @@ private void parallelAnnotation(JSONObject op, StringBuilder sb) {
         if (parallel != null && parallel) {
             sb.append("@parallel(width=");
             Object width = op.get("width");
-            if (width instanceof Integer)
+            if (width instanceof Integer) {
                 sb.append(Integer.toString((int) width));
+        	}
```
dlaboss: Why is it necessary / what's the benefit of adding code to also support a Long parallel width?  Integer is certainly large enough.

drhvm: It seems that the python width parameter has the type of long when written to the JSON.  I could not figure out how to force it to integer.
Using source line:
      topo = Topology("test_TopologyParallel")
      hw = topo.source(test_functions.hello_world)   
      hwp = hw.parallel(4)
      :

wmarshall484:
After merging .transform() and .lowLatency(), this needs to be brought up to date. @drhvm, could you rebase and from spl_python and push your changes to your fork?

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
ddebrunner: JSON4J reads any integral value as a long from serialized JSON.

However, parallel width is an integer, so the Java api should continue to be an int/Integer.

When reading the JSON, just do intValue() on the returned Number

ddebrunner:
Where possible I would encourage separate commits and pull requests for non-related fixes.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -83,6 +84,51 @@ def isolate(self):
...
+        Marks end of operators as parallel output with width
+        :returns: Stream
+        """
+        op = self.topology.graph.addOperator("$EndParallel$")
+        op.addInputPort(outputPort=self.oport)
+        oport = op.addOutputPort()
+        endP = Stream(self.topology, oport)
+        return endP.isolate()
+
+    def union(self, streamSet):
```
wmarshall484: Looks good, but I'm reticent to merge until the fix for the `removeOperator` bug has been delivered. Maybe this should be included in a separate pull request?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -83,6 +84,51 @@ def isolate(self):
...
+        op2.addInputPort(outputPort=iop.getOport())
+        oport = op2.addOutputPort(width)
+        return Stream(self.topology, oport)
+
+    def endParallel(self):
+        """
+        Marks end of operators as parallel output with width
+        :returns: Stream
+        """
+        op = self.topology.graph.addOperator("$EndParallel$")
```
wmarshall484: In the JAA, the endParallel code is as follows:
``` Java
    @Override
    public TStream<T> endParallel() {
        BOutput end = output();
        if(end instanceof BUnionOutput){
            end = builder().addPassThroughOperator(end);
        }
        end = builder().isolate(end);
        return addMatchingStream(builder().unparallel(end));
    }
```
There's a check which ensures that a $Union$ never immediately precedes an $EndParallel$ since it breaks the code generator. You should do something similar. If you see a $Union$ right before an $EndParallel$, add a "no-op" pass through operator as a buffer.

drhvm:
Please try running python tests before accepting.  I seem to be getting sigsegv's when the destructor runs for some of the operators (which I have not changed) Hopefully it is something in my envrionment.  Here is the stack I see when it fails:
     [exec] =====================================================================
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-apputils.so.0(Distillery::BacktraceDumper::dump(int, bool, int)+0x2b) [0x7f44f227a24b]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-runtime.so.0(Distillery::DistilleryApplication::faultHandler(int)+0x170) [0x7f44f40922b0]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-runtime.so.0(Distillery::DistilleryApplication::errorSignalHandler(int)+0x68) [0x7f44f4093158]
     [exec] /lib64/libc.so.6(+0x35670) [0x7f44f141d670]
     [exec] /usr/local/lib/libpython3.5m.so.1.0(+0xd2670) [0x7f44e12cf670]
     [exec] /usr/local/lib/libpython3.5m.so.1.0(+0x9ec57) [0x7f44e129bc57]
     [exec] /tmp/6528195857307019786/output/bin/test_TopologyIsolatedFilter.test_TopologyIsolatedFilter-a.dpe(SPL::_Operator::test_TopologyIsolatedFilter_OP2$OP::~test_TopologyIsolatedFilter_OP2$OP()+0x47) [0x7f44e173e977]
     [exec] /tmp/6528195857307019786/output/bin/test_TopologyIsolatedFilter.test_TopologyIsolatedFilter-a.dpe(SPL::_Operator::test_TopologyIsolatedFilter_OP2$OP::~test_TopologyIsolatedFilter_OP2$OP()+0x9) [0x7f44e173e9a9]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-spl-runtime.so.0(SPL::PEImpl::deleteOperators()+0x16b) [0x7f44f290109b]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-spl-runtime.so.0(+0x3b6e1f) [0x7f44f28fbe1f]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-spl-runtime.so.0(SPL::PEImpl::process()+0xf4) [0x7f44f29064a4]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-spl-runtime-standalone.so.0(SPL::StandaloneApplicationImpl::process()+0x176) [0x7f44f46f2326]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-spl-runtime-standalone.so.0(SPL::StandaloneApplicationImpl::run(std::vector<std::string, std::allocator<std::string> > const&)+0x101) [0x7f44f46f6d81]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-runtime.so.0(Distillery::DistilleryApplication::runSafeCatchDst(std::vector<std::string, std::allocator<std::string> > const&)+0x176) [0x7f44f40942f6]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-runtime.so.0(Distillery::DistilleryApplication::runSafeCatchall(std::vector<std::string, std::allocator<std::string> > const&)+0x13) [0x7f44f4091453]
     [exec] /home/sdickes/StreamsInstalls/V40/4.1.0.0/lib/libstreams-runtime.so.0(Distillery::DistilleryApplication::main(int, char**, char**)+0x3db) [0x7f44f389d42b]
     [exec] /lib64/libc.so.6(__libc_start_main+0xf5) [0x7f44f1409b15]
     [exec] /tmp/6528195857307019786/output/bin/standalone() [0x400ae1]

sdickes:
I think I found the bug. The code in the cpp dtors need to hold the GIL when doing Py_DECREF(). 

So in filter -- 
./com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
`// Destructor
MY_OPERATOR::~MY_OPERATOR()
{
    // Finalization code goes here
    if (function_) {
      Py_DECREF(function_);
    }

}
`
wouldn't we need to bracket this code with 

`// Destructor
MY_OPERATOR::~MY_OPERATOR()
{
  PyGILState_STATE gstate;
  gstate = PyGILState_Ensure();

    // Finalization code goes here
    if (function_) {
      Py_DECREF(function_);
    }

  PyGILState_Release(gstate);
}
`

I have done a quick little test and it works for me.

ddebrunner:
I just ran the ant test and everything passed, but it seems that maybe the python tests are not being run.

i thought they used to be run as part of the default test setup.

sdickes:
The api tests run under target test.application.api in test/python

ddebrunner:
Yeah, I could have sworn they were hooked into the main test target.

henrychi2:
Thanks for investigating the SEGV.  I'll start looking into a fix.  All the operators now have this problem.

henrychi2:
Created #290 for the SEGV.  Note if that gets merged, the two calls to PyGILState_Release(gstate) in PyFunctionSource_cpp.cgt from this pull request is no longer needed due to the RAII GIL lock class.

wmarshall484:
#290 has been merged. Please remove the unnecessary PyGilState_Release(gstate) commands.

drhvm:
Tests all work.  Should be ready to merge

test/python/topology/test1.py
```diff
@@ -110,6 +110,37 @@ def test_TopologySourceAndSinkCallable(self):
...
+  def test_TopologyParallelUnion(self):
+      topo = Topology("test_TopologyParallelUnion")
+      hw = topo.source(test_functions.hello_world)
+      hwp = hw.parallel(3)
+      hwf = hwp.filter(test_functions.filter)
+      hwf2 = hwp.filter(test_functions.filter)    
+      streamSet = {hwf2}
+      hwu = hwf.union(streamSet)
+      hwup = hwu.end_parallel()
+      hwup.sink(test_functions.check_union_hello_world)
```
wmarshall484: Since the source function is `test_functions.hello_world`, shouldn't the sink function be `test_functions.check_hello_world` instead of `test_functions.check_union_hello_world`?

test/python/topology/test1.py
```diff
@@ -110,6 +110,37 @@ def test_TopologySourceAndSinkCallable(self):
       topo = Topology("test_TopologySourceAndSinkCallable")
       hw = topo.source(test_functions.SourceTuplesAppendIndex(["a", "b", "c", "d"]))
       hw.sink(test_functions.CheckTuples(["a0", "b1", "c2", "d3"]))
+
```
wmarshall484: I don't see a test which actually validates that the data path was parallelized. For example, if you call `hw.parallel(4)`, there should be a test which asserts that the number of parallel channels == 4.

wmarshall484: The logic being, if parallel were implemented as follows:

``` python
# Effectively do nothing.
def parallel(self, num_channels):
    return self
```

These tests would still pass.

wmarshall484: One way we handled this in the JAA was by generating a unique token for each data channel, and then asserting that we had the correct number of unique tokens. E.g,
``` python
hw.parallel(4).generateUniqueToken().endParallel().countNumUniqueTokens()
```
and then assert that countNumUniqueTokens() outputs '4' when all the tuples have been processed.

wmarshall484:
+1, I'll follow up with an issue about additional .parallel() tests.

PR #274, opened 2016-02-26
Henrychi2 python spl multi transform

henrychi2:
initial drop for Python Application API multiTransform() support


ddebrunner:

ddebrunner:
Please do not change whitespace for the sake of it, it makes it harder to review pull requests as it has to be verified that the code that is not changing has in fact not been changed.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -72,6 +72,21 @@ def transform(self, func):
...
         return Stream(self.topology, oport)
+     
+    def multiTransform(self, func):
+        """
+        Transforms each tuple from this stream into 0 or more tuples using the supplied function.
+        For each tuple on this stream, the returned stream will contain all non-None tuples in
+        the iterator that is the result of the supplied function.
+        Tuples will be added to the returned stream in the order the iterator
+        returns them.
+        If the return is None or an empty iterator then no tuples are added to
```
ddebrunner: What happens if the return is not iterable?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -72,6 +72,21 @@ def transform(self, func):
         op.addInputPort(outputPort=self.oport)
         oport = op.addOutputPort()
         return Stream(self.topology, oport)
+     
+    def multiTransform(self, func):
+        """
+        Transforms each tuple from this stream into 0 or more tuples using the supplied function.
+        For each tuple on this stream, the returned stream will contain all non-None tuples in
```
ddebrunner: Should the function return an iterable or an iterator?

Or should it be either can be handled, would this approach be more Python-like?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_h.cgt
```diff
@@ -0,0 +1,40 @@
...
+  // Tuple processing for non-mutating ports
+  void process(Tuple const & tuple, uint32_t port);
+
+  // Punctuation processing
+  void process(Punctuation const & punct, uint32_t port);
+
+private:
+    // Members
+    PyObject *function_;
+    PyObject *iterableObjectFunction_;
```
ddebrunner: Why two functions, some comments here to indicate their purpose would help.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    // construct spl blob
+    long int size = PyBytes_Size(pyPickledReturnVar);
+    char * bytes = PyBytes_AsString(pyPickledReturnVar);          
+    pBlob.reset(new SPL::blob((const unsigned char *)bytes, size));
+    Py_DECREF(pyPickledReturnVar);
+            
+    // submit new tuple
+    OPort0Type otuple(*pBlob);
```
ddebrunner: We need to not hold the python lock (GILState) across tuple submission, I think that's what the other operators do.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
```diff
@@ -0,0 +1,187 @@
...
+      Py_DECREF(pyIterator);
+      break;
+    } else if(pyPickledReturnVar == 0){
+      PyErr_Print();
+      Py_DECREF(pyIterator);
+      PyGILState_Release(gstate);
+      throw;
+    }
+
+    // construct spl blob
```
ddebrunner: I created a set of utilities to map python objects to SPL tuples, any reason they are not being used?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
```diff
@@ -0,0 +1,187 @@
...
+  SPL::rstring value = ip.get_string();
+<%}%>
+
+  std::auto_ptr<SPL::blob> pBlob;
+  PyGILState_STATE gstate;
+  gstate = PyGILState_Ensure();
+
+  // invoke python application function and get return value
+  PyObject * pyBytes  = streamsx::topology::Splpy::pyBlobToBytes(value);
+  PyObject * pyReturnVar = streamsx::topology::Splpy::pyTupleFunc(function_, pyBytes);
```
ddebrunner: i think a follow on change could simplify this into a single function call into Python. That's what the other operators do so it would be good to follow that pattern. (see comment about two functions).

That is the call to get the iterator from the input tuple and to return the function that returns the pickled values can all be done in Python.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
henrychi2: Added the following comment for multiTransform():
The supplied function must return an iterable, otherwise a TypeError is raised.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
henrychi2: Yes, the function should return an iterable.  Added the following in the comment:
The supplied function must return an iterable, otherwise a TypeError is raised.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_h.cgt
henrychi2: Added comments in my latest commit.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
henrychi2: Fixed in my latest commit.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
henrychi2: As suggested, I merged the processing (depickling of the tuple, invoking the app function, creating an iterator, and pickling the return) into a single Python utility function.
The other function reference is just the application function, which along with the tuple, is needed as input to the utility function at process() time.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
henrychi2: Pickling of Python of objects results in blobs, so I am using PyBytes_Size() / PyBytes_AsString() to construct tuples.  This is consistent with how the source operator constructs SPL tuples from pickled values.
Other than the source operator, the only other place I see SPL tuples being created is py_functionReturnToTyples.cgt.  It is used in specific cases - the return type must be tuple or list, and it works with primitive times.  
The other functional operators like source, sink, filter work with tuples as blobs, so having multiTransform use blobs seemed appropriate.

test/python/topology/test_functions.py
```diff
@@ -86,4 +86,22 @@ def check_int_strings_transform_with_drop(t):
...
+   
+global expected_strings_multi_transform
+expected_strings_multi_transform = ["mary", "had", "a", "little", "lamb", "its", "fleece", "was", "white", "as", "snow"]
 
+def split_words(t):
+   return t.split()
+ 
+def check_strings_multi_transform(t):
+   print("TUPLE", t)
+   global _hwcount
```
wmarshall484: Just curious, there are a couple places in `test_functions.py` where you use the  `_hwcount` global variable. If you don't reset `_hwcount`, wouldn't it the tests that run afterwards?

henrychi2: Yes, I should probably use a different global variable to be safe.  I think it's with some luck that the test passes.  The test is run in standalone mode, so a new jvm is invoked, which means new python runtime, so globals are set to zero.  Also, the test case doesn't use other test functions that modify `_hwcount`.

PR #280, opened 2016-03-03
stateful operations using callable classes

henrychi2:
Stateful operations for filter(), transform(), multi_transform() via instances of callable classes

By allowing filter / transform function operators to accept instances of callable classes as input, state information such as user-defined parameters can be stored during class initialization and utilized when the class is called.

Miscellaneous fixes:
-Initialize function pointers to NULL in filter / transform custom operators
-Decrement function pointer reference count if necessary in operator destructor
-Additional error checking in custom operators
-Do not share `_hwcount` global variable among test functions


ddebrunner:
I looked quickly, but I didn't see anything explaining what the api is for the developer.

Is there an overview of what is being implemented here.

ddebrunner:

wmarshall484:
@henrychi2, maybe add a section in the spldoc like in pull request #283

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
```diff
@@ -130,13 +130,30 @@ void MY_OPERATOR::process(Tuple const & tuple, uint32_t port)
...
-  Py_INCREF(function_);
-  PyTuple_SetItem(funcArg, 0, function_);
-  // second argument is spl blob in bytes
+  // prepare to call the python utility function that invokes
+  // the application function and returns an iterator
+  PyObject * funcArg = PyTuple_New(3);
+  
+  // first argument is application function or serialized callable class
+  <% if ($serializedClass) { %>
+    PyTuple_SetItem(funcArg, 0, Py_BuildValue("s", <%=$serializedClass%>));
```
wmarshall484: Deserializing the callable class should be done once in the consturctor, not every time in the process method.

ddebrunner: Is this serializing a *class* or an *instance of a class* ?

I wouldn't expect we need to serialize a *class*.

wmarshall484: Instance of a class.

test/python/topology/test_functions.py
```diff
@@ -52,56 +52,112 @@ def string_to_int_except68(t):
...
+      self.upper = upper  
+   def __call__(self, tuple):
+      if len(tuple) > self.upper:
+         return True
+      return False
+
+global EXPECTED_STRINGS_LENGTH_FILTER
+EXPECTED_STRINGS_LENGTH_FILTER = ["goodbye", "farewell"]
+
+global _hwcount_length_filter
```
henrychi2: @wmarshall484 
Currently I define a global counter for each new testcase.  To avoid this, do you think it makes sense to support parameters for sink() as well (maybe source() too) ?  Thanks.

ddebrunner: No, the counters are a work-around for the fact that the testing infrastructure that the Java Application API supports is not yet supported by Python.

henrychi2:
Added some documentation and usage example to spldoc.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform_cpp.cgt
henrychi2: 1) Fixed so that deserializing the instance of the callable class is done once in the constructor.
2) Changed comments to state that we are serializing an instance of a class.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform.xml
```diff
@@ -49,13 +49,22 @@
...
-        <description>Function's name </description>
+        <description>Function or callable class's name</description>
         <optional>false</optional>
         <rewriteAllowed>true</rewriteAllowed>
         <expressionMode>AttributeFree</expressionMode>
         <type>rstring</type>
         <cardinality>1</cardinality>
       </parameter>
+      <parameter>
+        <name>serializedClass</name>
```
ddebrunner: So it's not a 'serializedClass' so can the name reflect what it represents.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/runtime.py
```diff
@@ -22,11 +23,24 @@ def _pickleReturn(v):
...
+# Given a callable object that is pickled and 
+# base64 encoded, deserialize and return the object
+def depickleCallable(serializedCallable):
+    return pickle.loads(base64.b64decode(serializedCallable))
+ 
+# Given a callable object F that is pickled and 
+# base64 encoded, return a function that
+# depickles the input and then calls F
+def depickleInputForCallable(serializedCallable) :
+    callableObject = pickle.loads(base64.b64decode(serializedCallable))
```
ddebrunner: Should this use the function declared above? `depickleCallable`

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -49,9 +49,14 @@ def sink(self, func):
 
     def filter(self, func):
         """
-        Filters tuples from a stream using the supplied function.
-        For each tuple on the stream the function is called passing
-        the tuple, if the function return evalulates to true the
+        Filters tuples from a stream using the supplied callable 'func'.
+        The callable input must be either 
+        1) the name of a top-level function defined using 'def' that takes a single parameter for the tuple, or
+        2) an instance of a top-level callable class that implements the method '__call__(self, tuple)'.  Using a
```
ddebrunner: What's a "top-level" function or class?

I don't see a standard definition in my python book.

I found this: https://docs.python.org/3.5/reference/toplevel_components.html
but it never defines what top-level means.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -62,10 +67,15 @@ def filter(self, func):
 
     def transform(self, func):
         """
-        Transforms each tuple from this stream into 0 or 1 tuples using the supplied function.
+        Transforms each tuple from this stream into 0 or 1 tuples using the supplied callable 'func'.
+        The callable input must be either 
+        1) the name of a top-level function defined using 'def' that takes a single parameter for the tuple, or
```
ddebrunner: Is there another way of defining functions that doesn't use 'def'?

ddebrunner:
At some point tests that have functions with mutable state are needed, can be a follow on pull request.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
wmarshall484: > the name of a top-level function defined using 'def'
How about "the name of a function defined in a module"?

wmarshall484:
I'd be curious to see whether you could pass a lambda to `filter` or `transform`. As far as I know, lambdas are callable objects, so they might work with the current code.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionMultiTransform/PyFunctionMultiTransform.xml
henrychi2: Change parameter and variable names to 'serializedCallableInstance'.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/runtime.py
henrychi2: Fixed in latest commit.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
henrychi2: Changed the comment to read "functions defined at the top level of a module".  This is the same terminology used in Python doc here:
https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled

All I want to say here is that we don't support nested functions, since an inner function is local the outer function and thus not visible by name to PythonC.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
henrychi2: I guess lambda functions can also be 'defined', which we don't support.
Removed the comment 'using def', since we already mention that we don't support lambdas in the spldoc.

henrychi2:
As a sanity check, I added a testcase with mutable state:

```
class IncMaxSplitWords:
   def __init__(self, maxsplit):
      self.maxsplit = maxsplit  
   def __call__(self, tuple):
      words = tuple.split(None, self.maxsplit)
      # test mutable state, increase maxsplit by 1 for next tuple
      self.maxsplit += 1
      return words
```

henrychi2:
Unfortunately it doesn't seem like lambdas work with `filter` or `transform` in the current implementation.  This is because lambdas are anonymous, so in PythonC we don't have a reference to a function name.
`Fatal error: function not callable: <lambda>`

Another way is to somehow serialize the lambda.  But pickle doesn't support this.  Doing some search on the web, lambdas might be serialized with marshal, but it might have limitations.  Or lambas can be serialized with third-party libraries like dill.  This will require more investigation in the future.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionTransform/PyFunctionTransform_cpp.cgt
```diff
@@ -62,14 +64,27 @@ MY_OPERATOR::MY_OPERATOR()
...
-      streamsx::topology::Splpy::loadFunction("streamsx.topology.runtime", "depickleInput");
-
+    PyObject * depickleInput = NULL;
     PyObject * funcArg = PyTuple_New(1);
-    PyTuple_SetItem(funcArg, 0, appFunction);
+    
+    <% if ($serializedCallableInstance) { %>
+      depickleInput = streamsx::topology::Splpy::loadFunction("streamsx.topology.runtime", "depickleInputForCallableInstance");
+      PyTuple_SetItem(funcArg, 0, Py_BuildValue("s", <%=$serializedCallableInstance%>));
+      Py_DECREF(appFunction);
```
ddebrunner: I found trying to understand the use of appFunction hard, it used to be the function that was going to be called (or wrapped by another function), but what is appFunction in the case it's a callable class? What is functionName set to when the function is loaded at line 62?

ddebrunner:
I think the code would become clearer if the operator parameter names reflected what they are, rather than what they used to be.

`functionName --> name` ??? Maybe something better here. `pyName` ??

`functionModule --> module`

`serializedCallableInstance --> `callable` - Just easier to talk about than something with 7 syllables .

The current code is hard to understand because 'functionName' represents a function name or a class name, but a reader naturally assumes it represents only a function name.

ddebrunner:
Will there be a follow pull request for Sink and Source?

henrychi2:
@ddebrunner 
1) Sure, I will change the operator parameter names.
2) Just want to confirm...we want Sink and Source to accept instances of callable classes correct?  Let me know if I misunderstood, otherwise I can start working on that and include in this pull request.

ddebrunner:
I think Sink/Source can be a follow on pull request, let's get this one merged first.

henrychi2:
Changed the operator parameter names with latest commit:

`functionName --> pyName`
`functionModule --> pyModule`
`serializedCallableInstance --> pyCallable`

PR #282, opened 2016-03-03
Fix for GraphUtilities::removeUnionOperator not finding all upstream …

drhvm:
This is the fix required for python Union to work correctly.  I was not able to create a specific test in java for this condition but my unittests seem to run

### DIFF
--- a/java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
@@ -196,13 +196,13 @@ public class GraphUtilities {
                 }
             }
             
-            // Also get input name
-            String operatorInName="";
+            // Also get input names
+            List<String> operatorInNames = new ArrayList<String>();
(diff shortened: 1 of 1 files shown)


ddebrunner:

wmarshall484:
To verify this works, uncomment lines 189 and 190 in com.ibm.streamsx.topology.test.api.IsolateTest, and verify that the test passes. It seems for some reason that this test (which would have caught the error) was not being verified.

wmarshall484:
+1 When com.ibm.streamsx.topology.test.api.IsolateTest#unionIsolateTest passes.

drhvm:
I am having no luck getting that IsolateTest to run with the asserts uncommented, either with my fix or without.  I don't see any other tests with StreamsContext.Type.TOOLKIT that use the assertTrue method call.  Is this valid in the test framework?

drhvm:
I was able to check that the expectedCount which was supposed to be 4 was actually 0.  The json output looks correct.

PR #287, opened 2016-03-09
Henrychi2 python stateful operations for Source and Sink

henrychi2:
Stateful operations for source() and sink() via instances of callable classes

By allowing Source and Sink operators to accept instances of callable classes as input, state information such as user-defined parameters can be stored during class initialization and utilized when the class is called.


ddebrunner:

ddebrunner:
I'll merge this but I think a common theme across the callable code is that I'm not sure we need a depickleInputForCallableInstance.

Seems like within Python we should be able to figure out the right thing to do in depickleInput so the the cpp.cgt code is simpler for all of these operators.

I'll take a look at that separately.

ddebrunner:
Just to add the reason for trying to have a single function is to try and keep the cgt file simple, otherwise as more callable types may be supported (e.g. lambdas) the cgt will become too complicated if we have many depickleInputForXXX functions.

PR #288, opened 2016-03-11
update to low_latency and end_low_latency

sdickes:
update the function names from lowlatency/endlowlatency to low_latency/end_low_latency


ddebrunner:

PR #290, opened 2016-03-12
Henrychi2 python spl destructor segv

henrychi2:
segv due to Py_DECREF call in operator destructor without GIL lock

Use RAII class to lock and unlock the GIL


ddebrunner:

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionTransform/PyFunctionTransform_cpp.cgt
```diff
@@ -99,13 +97,13 @@ MY_OPERATOR::MY_OPERATOR() : function_(NULL), pickleObjectFunction_(NULL)
...
 <%}%>
 
-    PyGILState_Release(gstate);
 }
 
 // Destructor
 MY_OPERATOR::~MY_OPERATOR() 
 {
     // Finalization code goes here
+    streamsx::topology::PyGILLock lock;
```
ddebrunner: Why do we get the lock when function is null?

henrychi2: Fixed in latest commit so that lock is not acquired when function is null.

Issue #294, opened 2016-03-21
[Python] [enhancement] importing application modules and packages


henrychi2:
**Overview**
Operations in the Python API such as filter() and transform() accept a user-supplied function as input parameter.  The function is defined in a module created by an application developer.  To allow the application to be executed from different locations, dependencies such as these application modules are copied into the Streams application bundle (.sab file).

In the current implementation, when the Python topology is submitted and compiled into a bundle, each module containing user-supplied functions is automatically copied from the local file system to a target directory in the bundle (`<toolkitRoot>/opt/python/modules`).  When the bundle is executed, the target directory is automatically added to the Python search path `sys.path`, allowing the modules to be imported and its functions to be called at runtime.

**Limitations**
There are a couple of limitations in the current implementation.

1)
Imported modules are not allowed to be defined as part of a package.  When individual modules are copied into the application bundle, the directory structure is not maintained.  For example,
```
test1.py
test_package/  <== package
    __init__.py
    test_module.py  <== user-supplied functions in a module inside a package
```
```
test1.py
from test_package import test_module
...
     topo = Topology("test_TopologyImport")
     source = topo.source(test_module.test_func) 
```
For source(), the input function's module name is actually `test_package.test_module`.  In the .sab, we see that test_module.py is copied without the package hierarchy.
```
<toolkitRoot>/opt/python/modules/
    test_module.py
```
The runtime tries to import the module named `test_package.test_module` but cannot find it.  The application fails execution with error:
`ImportError: No module named 'test_package'`

2)
If the imported module imports another module, execution might fail with `ImportError`.  This is because only the module containing the function used as input to the operation gets copied into the application bundle.  The module's dependencies are not.

**Possible Solutions**

1)
To support importing of application packages, we could have the Python API detect if the module containing user-supplied functions is part of a package.  If so, automatically copy the package directory from the local file system to a target destination in the bundle (`<toolkitRoot>/opt/python/packages`).  This target destination is also added to the Python search path, allowing modules in the package to be imported at runtime.

2)
To support importing of a module's dependencies, we could introduce a new method that lets the application add a module file or package directory to the application runtime.  For example,
```
# Adds a Python module or package to the application runtime
# If the path is for a Python module file, the file is copied to <toolkitRoot>/opt/python/modules
# If the path is for a Python package, the package is copied to <toolkitRoot>/opt/python/packages
Topology.add_dependency(self, path)
```
**Edit 2: Summary of proposed changes**

1)
[...]

PR #297, opened 2016-03-24
Henrychi2 python spl import packages

henrychi2:
* Support importing input functions from modules in packages
* Support importing input functions from modules that imports other modules
* Delete splpytmp json file from /tmp after _submitUsingJava()


ddebrunner:

ddebrunner:
There doesn't seem to be a summary of what the proposed solution was in #294 or the wiki page that was created.

henrychi2:
Added a section, "Summary of proposed changes", to the first post in #294.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -157,8 +165,7 @@ def _addOperatorFunction(self, function):
             # pickle format is binary; base64 encode so it is json serializable 
             self.params["pyCallable"] = base64.b64encode(pickle.dumps(function)).decode("ascii")
 
-        self.params["pyModule"] = function.__module__
-                    
+        self.params["pyModule"] = streamsx.topology.dependency.get_module_name(function)
```
ddebrunner: Maybe a comment here why the simpler ` function.__module__` cannot be used.

java/src/com/ibm/streamsx/topology/internal/context/ToolkitStreamsContext.java
```diff
@@ -232,7 +232,8 @@ public FileVisitResult preVisitDirectory(final Path dir,
             public FileVisitResult visitFile(final Path file,
                     final BasicFileAttributes attrs) throws IOException {
                 Files.copy(file,
-                        targetPath.resolve(sourcePath.relativize(file)));
+                        targetPath.resolve(sourcePath.relativize(file)),
+                        StandardCopyOption.REPLACE_EXISTING);
```
ddebrunner: Why is REPLACE_EXISTING needed?

It seems to imply that the same file is being copied multiple times.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
```diff
@@ -0,0 +1,173 @@
...
+                imported_modules[val.__name__] = val
+    return imported_modules
+
+def is_builtin_module(module):
+    return module.__name__ in sys.builtin_module_names and \
+           not hasattr(module, '__file__') and \
+           not hasattr(module, '__path__')
+
+def _is_streamsx_module(module_path):
+    return "com.ibm.streamsx.topology" in module_path
```
ddebrunner: Why not just check the module name starts with streamsx.topology?

Using the path seems dubious.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
```diff
@@ -0,0 +1,173 @@
...
+
+def is_streamsx_module(module):
```
ddebrunner: Seems over complex, like the previous comment, just check it's streamsx.topolology.*

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
henrychi2: Added a comment.  Also pulled out the change since we are no longer supporting functions from the `__main__` module as input to operations in this pull request.

java/src/com/ibm/streamsx/topology/internal/context/ToolkitStreamsContext.java
henrychi2: Due to namespace packages, two directories can be merged (both copied into `opt/python/packages).  This change is needed if two directories contain a conflicting file name.  Obviously only one file can exist in the target destination; the other is overwritten.  The file that wins should be the module that is actually loaded (comes first in the Python search path).

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
henrychi2: Fixed in latest commit.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
henrychi2: Fixed in latest commit.

ddebrunner:
Merging this to allow follow on focused pull requests to address any issues.

PR #298, opened 2016-04-01
[WIP] Subcribe support for python receiving rstring SPL type

sdickes:
WIP
subscribe support for python receiving rstring SPL type
rstring is converted to python string, pickled, and sent downstreams

### DIFF
--- a/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionSubscribeString/PyFunctionSubscribeString.xml
+++ b/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionSubscribeString/PyFunctionSubscribeString.xml
@@ -1,0 +1,93 @@
+<?xml version="1.0" ?>
+<operatorModel
+  xmlns="http://www.ibm.com/xmlns/prod/streams/spl/operator" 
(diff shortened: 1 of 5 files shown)


ddebrunner:
Could you expand more of what the user functionality is?

I wouldn't have expected rstrings to be pickled.

ddebrunner:

sdickes:
What I added was for python to subscribe to a topic that has type rstring. The input type is rstring so a new C++ operator type is inserted into the graph to: convert the rstring to a Python string; pickle the result; and then submit the new tuple to the output port. This allows the next operator, written in Python, to process this input tuple. 

This is a [WIP] as I am not sure if this is the correct approach but it definitely works. 
A fuller question is what types should Python support and is there a generic approach to support them?

ddebrunner:
My intention (and some fragments of code exist in the operators) that the rstring for the Schema String would be passed directly into the Python code (as a Python string) as the function argument.
No need to create an additional operator to pickle it.

sdickes:
So the fragments are the $attr checks for '__spl_po' or 'string' or 'jsonString' that set $pystyle?

ddebrunner:
yes

sdickes:
Tthis pull request should be closed based on the other pubsub changes, correct?

Issue #292, opened 2016-03-16
Add additional tests for parallel()


wmarshall484:
See comments in PR #272

PR #299, opened 2016-04-06
Drh parallelhash

drhvm:
Additional code to support all routing types on parallel operator.
Modified and additional tests for all routing types   ##included.
Resolves issue #292


ddebrunner:

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -184,6 +184,37 @@ namespace streamsx {
...
+    static std::auto_ptr<SPL::int32> pyTupleHash(PyObject * function, PyObject * pickleObjectFunction, SPL::blob & pyblob) {
+
+      std::auto_ptr<SPL::int32> ret;
+      PyGILLock lock;
+
+      // convert spl blob to bytes
+      PyObject * pyBytes  = pyBlobToBytes(pyblob);
+      // invoke python nested function that calls the application function
+      PyObject * pyReturnVar = pyTupleFunc(function, pyBytes);
+      if (pyReturnVar == Py_None){
```
ddebrunner: Could we have some comments explaining what is going on?

I would have thought getting a hash was a simple situation, every object has a hash but there is a three way return from some python function, 0 None or something else.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionHashRemover/PyFunctionHashRemover.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" ?>
+<operatorModel
```
ddebrunner: I don't understand why we need a new operator to remove an attribute. Just have a Filter with a different schema on the output from the input.

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -184,6 +184,37 @@ namespace streamsx {
...
+        Py_DECREF(pyReturnVar);
+        return ret;
+      } else if(pyReturnVar == 0){
```
ddebrunner: Isn't 0 a valid hash? Guess I'm missing something here.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionHashRemover/PyFunctionHashRemover.xml
drhvm: I assume you mean a Python Filter?  Not sure I get why this is allowed since the PyFunctionFilter xml of the output port says: <tupleMutationAllowed>false</tupleMutationAllowed>
Am I missing something?

drhvm: Sorry....output port xml says tupleMutationAllowed=false

ddebrunner: No, SPL standard toolkit Filter.

tupleMutationAllowed=false is unrelated, it means that downstream fiused operators are not allowed to modify the tuple reference they receive.

drhvm: When I try to use spl.relational::Filter as the hash remover and set the output port schema to remove the hash attribute, I get the following compile error:
CDISP9083E ERROR: The schema at output port number 0 must match the schema at input port number 0.
Which seems to match the doc that says:
The Filter operator requires that the stream type of the output port matches the stream type of the input port. 
What am I missing?

ddebrunner: Sorry, didn't think a Filter required matching schemas, then Functor will work.

drhvm:
Latest commit resolves issue #329

drhvm:
Pushed up the changes from the latest merge from python_spl

com.ibm.streamsx.topology/com.ibm.streamsx.topology.python/namespace-info.spl
```diff
@@ -454,27 +463,28 @@ The sample `transform_sample.py` can be found under `samples/python/topology/sim
...
+* `STANDALONE` converts the application to an SPL graph, compiles it and executes it as an IBM Streams standalone application.  The standalone execution is spawned as a separate process.
+
 Limited testing has been performed to build an application using DISTRIBUTED and BUNDLE.
 
 # Future Items
 
 Future items that need to be supported
 
-* `parallel()` function on `Stream` utilizing hashed distribution
+* DISTRIBUTED submission
```
ddebrunner: Remove DISTRIBUTED

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionHashAdder/PyFunctionHashAdder_cpp.cgt
```diff
@@ -0,0 +1,168 @@
...
+<% if ($pystyle eq 'python') { %>
+  SPL::blob value = ip.get___spl_po();
+<%}%>
+
+<% if ($pystyle eq 'string') { %>
+  SPL::rstring value = ip.get_string();
+<%}%>
+
+
+  std::auto_ptr<SPL::int32> spl_hash = streamsx::topology::Splpy::pyTupleHash(function_, pickleObjectFunction_, value);
```
ddebrunner: Just curious why we need the auto_ptr?

Wouldn't it be simpler to just have pyTupleHash return an int and assign it directly to the tuple?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionHashAdder/PyFunctionHashAdder_cpp.cgt
```diff
@@ -0,0 +1,168 @@
...
+      PyErr_Print();
+      Py_DECREF(depickleInput);
+      Py_DECREF(funcArg);
+      throw;
+    }
+
+    Py_DECREF(depickleInput);
+    Py_DECREF(funcArg);
+    
+    pickleObjectFunction_ =
```
ddebrunner: Why is this needed, there's no pickling going on, or shouldn't be.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionHashAdder/PyFunctionHashAdder_cpp.cgt
```diff
@@ -0,0 +1,168 @@
...
+    appDirSetup += "\")\n";
+
+    const char* spl_setup_appdir = appDirSetup.c_str();
+    if (PyRun_SimpleString(spl_setup_appdir) != 0) {
+         SPLAPPTRC(L_ERROR, "Python script splpy_setup.py failed!", "python");
+         PyErr_Print();
+         throw;
+    }
+
+<%
```
ddebrunner: Since the pr was first created the Python functional operators have been simplified to use common code.

Any chance of having this one align with the others?

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -188,6 +188,30 @@ namespace streamsx {
...
+     
+       // construct integer from  return value
+      if(PyLong_Check(pyReturnVar)) {
+        SPL::int32 retval = PyLong_AsLong(pyReturnVar);          
+        ret.reset( new SPL::int32(retval));
+        Py_DECREF(pyReturnVar);		
+        return ret;
+      } else {
+		Py_DECREF(pyReturnVar);
+		PyErr_Print();
```
ddebrunner: Need to use the new function I added the does PyErr_Print and then flushes.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
```diff
@@ -24,6 +24,7 @@ class CommonSchema(enum.Enum):
     Json = StreamSchema("tuple<rstring jsonString>")
     String = StreamSchema("tuple<rstring string>")
     Binary = StreamSchema("tuple<blob binary>")
+    PythonHash = StreamSchema("tuple<blob __spl_po,int32 __spl_hash>")
```
ddebrunner: Please move elsewhere, `PythonHash` is not a common interchange schema.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.python/namespace-info.spl
```diff
@@ -356,38 +356,45 @@ To avoid name conflicts, do not create modules with the same name as
...
      
-  param `width`: degree of parallelism
+  param `width`: degree of parallelism
+
+  param: `routing` - denotes what type of tuple routing to use. 
+  * ROUND_ROBIN: delivers tuples in round robin fashion to downstream operators (Default)
+  * HASH_PARTIONED: delivers to downstream operators based on the hash of the tuples being sent
+  * KEY_PARTIONED: delivers to downstream operators based on a hash provided by the user function provided
+  
+  param: `func` - User defined function called when KEY_PARTIONED routing is specified. The function provides an int32 value to be used as the hash that determines the tuple routing to downstream operators
```
ddebrunner: Just a function right, not 'user defined function'

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/internal/functional/ops/PyInternalFunctions.py
```diff
@@ -0,0 +1,3 @@
+def hashTuple(t) :
```
ddebrunner: No need to create a new file, just use the existing runtime.py

com.ibm.streamsx.topology/com.ibm.streamsx.topology.python/namespace-info.spl
```diff
@@ -356,38 +356,45 @@ To avoid name conflicts, do not create modules with the same name as
...
   as the calling stream.
   
   return: Stream
 
 * `union(self, streamSet)` function on `Stream`
 
-  Creates a stream that is a union of this stream and other streams
-
-  param `streamSet`: a set of Stream objects to merge with this stream
+  The Union operator merges the outputs of the streams in the set into a single stream.
```
ddebrunner: merge error

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -232,15 +235,45 @@ def parallel(self, width):
         
         Args:
             width (int): degree of parallelism
+            routing - denotes what type of tuple routing to use. 
+                ROUND_ROBIN: delivers tuples in round robin fashion to downstream operators
+                HASH_PARTIONED: delivers to downstream operators based on the hash of the tuples being sent
+                KEY_PARTIONED: delivers to downstream operators based on a hash provided by the user function provided
+            func - User defined function called when KEY_PARTIONED routing is specified.  The function provides an
```
ddebrunner: This is different to how the Java and SPL define parallel, where the key is defined, not the hash on a key.

Not saying this is wrong, but what's the reason for it being different. Really, the routing type is no longer KEY_PARTITIONED, but HASH_PARTITIONED with an application supplied hash.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
drhvm: Any suggestions on where to put it?  Should I just create a new class ?
class InternalSchema(enum.Enum):
    """Internal stream schemas"""
    PythonHash = StreamSchema("tuple<blob __spl_po,int32 __spl_hash>")

    def schema(self):
        return self.value.schema();

    def spl_json(self):
        return self.value.spl_json()

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/internal/functional/ops/PyInternalFunctions.py
ddebrunner: Is this function used by the runtime or is it just a test function?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -232,15 +235,45 @@ def parallel(self, width):
...
-        return Stream(self.topology, oport)
+        if (routing == None or routing == Routing.ROUND_ROBIN) :
+            iop = self.isolate()                  
+            op2 = self.topology.graph.addOperator("$Parallel$")
+            op2.addInputPort(outputPort=iop.getOport())
+            oport = op2.addOutputPort(width)
+            return Stream(self.topology, oport)
+        elif(routing == Routing.HASH_PARTITIONED or routing == Routing.KEY_PARTITIONED) :
+            if (routing == Routing.HASH_PARTITIONED) :
+                op = self.topology.graph.addOperator("com.ibm.streamsx.topology.functional.python::PyFunctionHashAdder", PyInternalFunctions.hashTuple)
```
ddebrunner: Do we need hashTuple? Why can't I just pass hash  (the builtin one)?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
ddebrunner: Since it's only used in one place why not just use Schema directly?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
drhvm: You are right....I guess I am not clear on what the design point is for KEY_PARTITIONED.  How about I remove KEY_PARTITIONED for now and just include HASH_PARTITIONED with an optional user provided hash.  Then we can figure out what we really want for KEY_PARTITIONED

ddebrunner: We can remove it for now.

Basically it will need a function returning a key and then the runtime guarantees that all tuples with the same key are sent to the same channel. It will use the same mechanism, basically we would wrap the user function that returned the key to return  hash(key), or hash(userFunction(tuple))

drhvm:
I have made the suggested changes above and also removed the KEY_PARTITION option to be added later.

drhvm:
Latest update resolves issue #329

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -232,15 +234,43 @@ def parallel(self, width):
...
-        return Stream(self.topology, oport)
+        if (routing == None or routing == Routing.ROUND_ROBIN) :
+            iop = self.isolate()                  
+            op2 = self.topology.graph.addOperator("$Parallel$")
+            op2.addInputPort(outputPort=iop.getOport())
+            oport = op2.addOutputPort(width)
+            return Stream(self.topology, oport)
+        elif(routing == Routing.HASH_PARTITIONED ) :
+            if (func is None) :
+                op = self.topology.graph.addOperator("com.ibm.streamsx.topology.functional.python::PyFunctionHashAdder", hash)
```
ddebrunner: Minor point but this can be simplified to just set func to hash when func is None and then no need to repeat the addOperator line.

drhvm: good point....just pushed update

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -188,6 +188,28 @@ namespace streamsx {
...
+
+      PyGILLock lock;
+
+      // convert spl blob to bytes
+      PyObject * pyBytes  = pyBlobToBytes(pyblob);
+      // invoke python nested function that generates the int32 hash
+      PyObject * pyReturnVar = pyTupleFunc(function, pyBytes); 
+     
+       // construct integer from  return value
+      if(PyLong_Check(pyReturnVar)) {
```
ddebrunner: I don't think we need this check. We just call PyLong_AsLong()

Issue #309, opened 2016-04-08
[Python] Depickle errors.


ddebrunner:
Seeing corrupted tuple (blob) before depickling.

Originally the source operator created the blob before the reference was decremented:
https://github.com/IBMStreams/streamsx.topology/blob/7e79a1a07311ec973793187f37b4fc28f62b3490/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionSource/PyFunctionSource_cpp.cgt#L100

Now it is after:

https://github.com/IBMStreams/streamsx.topology/blob/python_spl/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionSource/PyFunctionSource_cpp.cgt#L125

The issue is that the return from PyBytesAsString is a pointer into the Python bytes object, not a copy, so we need to keep the reference count until the SPL Blob constructor has done its copy.

PR #311, opened 2016-04-09
[must-fix for alpha] Henrychi2 python spl corrupted tuple in source operator

henrychi2:
Fix for issue #309 - corrupted tuple (blob) in source operator


ddebrunner:

ddebrunner:
I created #312 which seems to be a simpler & clearer approach, no need for an auto_ptr.

Any idea which is preferable?

henrychi2:
Yes, your fix is simpler and clearer.  Closing mine...thanks.

Issue #304, opened 2016-04-08
[Python] Add Sample transform application to samples.


ddebrunner:
The Python Application API has no samples, would be good to add this one and then also reference it from the SPLDOC.

henrychi2:
Must-fix for alpha

henrychi2:
Adding a sample, and updating spldoc to reference its location

Issue #303, opened 2016-04-08
[Python] Sample transform has confusing isolates.


ddebrunner:
The use of isolates seem somewhat arbitrary, there's no comment as to why they are there.

Can lead to others copying them blindly as that's what the sample does.

Maybe just remove them.

henrychi2:
Must-fix for alpha

henrychi2:
Removing isolates for now, since this is a basic sample to get started.

Issue #301, opened 2016-04-08
[Python] SPLDOC says only submission to STANDALONE is supported.


ddebrunner:
Submitting as a BUNDLE seems to work as well.

henrychi2:
Must-fix for alpha

henrychi2:
Updating the spldoc to say that the BUNDLE context is supported

Issue #300, opened 2016-04-08
[Python] Sample Transform example missing mymodule.print


ddebrunner:
In the SPLDOC the Sample Transform Application includes this:

`iso2.sink(mymodule.print)`

but this function does not exist below like string_to_int does.

It works just using print

`iso2.sink(print)`

henrychi2:
Must-fix for alpha

henrychi2:
Changing sample to use the built-in print() function

PR #314, opened 2016-04-09
[alpha] updates to spldoc and sample (issues #300,  #301, #303, #304)

henrychi2:
Updates to Python spldoc, plus adding a sample (issues #300,  #301, #303, #304)


ddebrunner:

Issue #308, opened 2016-04-08
[Python] Pip installed packages are not copied into the bundle.


ddebrunner:
I tried an example using TwitterAPI where there were two "application modules"

streamsapp.py - contained the topology calls and imported streamsmodules to use a callable

streamsmodules.py - contained the callable and imported TwitterAPI

Running the bundle has TwitterAPI module not found.

ddebrunner:
Not clear from the SPLDOC what should happen in this case.

henrychi2:
Must-fix for alpha

henrychi2:
I was able to install TwitterAPI and recreate the issue.  With changes in pull request #313, I manually inspected the .sab and can see that TwitterAPI and other dependent packages are copied.

ddebrunner:
Verified, with #313 and #319 the twitter api and dependencies were copied into the bundle, actually 297 extra files were included.

I ran a standalone bundle:
* without PYTHONPATH being set
* and the com.ibm.streamsx.topology toolkit removed (to ensure there was no hard-coded path dependency)

and it worked.

PR #319, opened 2016-04-10
[alpha] henrychi2 python spl dependency fixes2

henrychi2:
1) A module dependency is not copied to the bundle if a dependent function is imported
Works:
import my_module
Doesn't work:
from my_module import my_func

Modified an existing testcase to capture this case

2) Changed the logic to determine if a path is for a site-package.  Packages installed in `site-packages` should be copied to the bundle.

These changes may be related to #308.


ddebrunner:

Issue #323, opened 2016-04-11
[Python] Module not found running in distributed.


ddebrunner:
Running an app with tweepy worked fine in standalone but distributed had an error loading my module (`tweets.py`) holding the functions I cal from my topology. It fails trying to find the module 'six' which is referenced from the tweepy package I import from `tweets.py`.

Looking at the files in the sab six.py was included (hence discovered by the dependency checker) but it was put into a location that is not on the module path.

`opt/python/packages/requests/packages/urllib3/packages/six.py`

henrychi2:
Must-fix for alpha

henrychi2:
There is another `six.py` that should have been copied but wasn't (`opt/python/packages/six.py`).  `six.py` is a package that is not a directory but a single file, which I didn't account for.  I have potential fix with pull request #325.

ddebrunner:
tweepy sample works!

PR #325, opened 2016-04-11
[alpha] Henrychi2 python spl dependency fixes3 (issue #323)

henrychi2:
Dependency resolve is not handling the case where a module has an empty `__path__`

This is a potential fix for issue #323


ddebrunner:

PR #333, opened 2016-04-12
set executable permissions on pyversion.sh in build target create_rel…

sdickes:
During testing I discovered that the pyversion.sh file does not have the "x" permission in the release tarfile. This looks to be normal behavior for ant so I updated the "create_release_bundle" target tar  step in the root build.xml to do the following -- 

set permissions to 755 on pyversion.sh
include the files previously included 
exclude the files previously 
exclude pyversion.sh because it was added to the tar file twice otherwise

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -102,10 +102,17 @@
       <property name="release.dir" location="${release.base}" />
       <mkdir dir="${release.dir}" />
       <property name="releasefilename" value="${release.dir}/streamsx.topology-v1.2-${DSTAMP}-${TSTAMP}.tgz"/>
-      <tar compression="gzip" longfile="gnu" destfile="${releasefilename}"
-         basedir="${basedir}"
-         includes="com.ibm.streamsx.topology/** samples/**"
-         excludes="**/.gitignore" />
+      <tar compression="gzip" longfile="gnu" destfile="${releasefilename}">
+         <tarfileset dir="${basedir}" filemode="755" >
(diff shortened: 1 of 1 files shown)


sdickes:
must fix for the alpha

ddebrunner:

build.xml
```diff
@@ -102,10 +102,17 @@
       <property name="release.dir" location="${release.base}" />
       <mkdir dir="${release.dir}" />
       <property name="releasefilename" value="${release.dir}/streamsx.topology-v1.2-${DSTAMP}-${TSTAMP}.tgz"/>
-      <tar compression="gzip" longfile="gnu" destfile="${releasefilename}"
-         basedir="${basedir}"
-         includes="com.ibm.streamsx.topology/** samples/**"
-         excludes="**/.gitignore" />
+      <tar compression="gzip" longfile="gnu" destfile="${releasefilename}">
+         <tarfileset dir="${basedir}" filemode="755" >
```
ddebrunner: Does it need to be writeable?

sdickes:
I originally made it 555 but saw other files in the install tree were also writable by the owner. So for this change I made it writable to match other files. For instance: 

24663637    4 -rw-r--r--   1 sdickes  streams      2542 Apr 11 09:53 ./com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt

PR #335, opened 2016-04-13
Add python samples

henrychi2:
Adding python samples.  I just ported the ones I could from Java samples.

I also added the temperature sensor sample from the application development guide.


ddebrunner:

samples/python/topology/parallel/parallel_regex_grep_functions.py
```diff
@@ -0,0 +1,83 @@
+import time
+import os
+import re
+import logging
+
+logging.basicConfig(level = logging.INFO, format = "%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+logger = logging.getLogger(__name__)
+
+class DirectoryWatcher:
```
ddebrunner: Maybe a common function module rather than duplicate code?

samples/python/topology/games/fizz_buzz_functions.py
```diff
@@ -0,0 +1,35 @@
+import time
+
+def int_tuples():
+    """
+    Generator function that returns a integer count 
+    with each iteration (up to 300)
+    """
+    for count in range(300):
```
ddebrunner: I like this, Python is powerful

samples/python/topology/parallel/parallel_regex_grep_functions.py
henrychi2: Created a common module, util_functions.py, to avoid duplicate code.

Issue #339, opened 2016-04-14
[Python] No need to check for long type when hashing


ddebrunner:
See comment in pr #299

PR #340, opened 2016-04-14
fix for issue #339

drhvm:
Catching exception to print and rethrow


ddebrunner:

PR #343, opened 2016-04-14
create synonyms for transform (map) and multi_transform (flat_map)

henrychi2:
* Create synonyms for transform (map) and multi_transform (flat_map).  

Apache Quarks has Map and FlatMap functions instead of Transform and MultiTransform as they are more widely used terms.

Keep transform() and multi_transform() functions, to be consistent with the Java Topology API.

* Fix a minor issue in the grep.py sample, to skip binary files


ddebrunner:

PR #350, opened 2016-04-20
update python samples

henrychi2:
1) Add hash-partitioned parallel sample (ported from Java sample)
2) Update grep sample to skip subdirectory entries when searching for files to read in a directory
3) Change logic for computing arithmetic mean and standard deviation in outlier sample to avoid keeping all tuples in memory

### DIFF
--- a/samples/python/topology/state/find_outliers_functions.py
+++ b/samples/python/topology/state/find_outliers_functions.py
@@ -1,5 +1,5 @@
 import random
-import statistics
+import math
 
 def readings():
     """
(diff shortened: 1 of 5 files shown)


ddebrunner:

samples/python/topology/state/find_outliers_functions.py
```diff
@@ -1,5 +1,5 @@
 import random
-import statistics
+import math
```
ddebrunner: Why the change in packages?

henrychi2: The `statistics` package has functions to compute the mean and stddev given a list of tuples.  Since we don't want to keep a history of all tuples in memory, we are changing the logic to compute the mean and stddev incrementally.  This requires some functions in the `math` library (like math.sqrt()), and the `statistics` package is no longer needed.
Computing the mean and stddev incrementally is already done by the Java sample.  I didn't do a proper port with the first delivery of the Python sample.

Issue #286, opened 2016-03-09
[Python] Have consistent naming for functions.


ddebrunner:
Looking the at the SPLDOC describing the python Streams functions we have inconsistent naming for functions:

multi_transform
lowlatency
endlowlatency

PEP-0008 states:
> Function names should be lowercase, with words separated by underscores as necessary to improve readability.

https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

sdickes:
makes sense, I will update the latency functions to use "_"  and I think @drhvm has the same problem with endparallel.

sdickes:
lowlatency and endlowlatency have been updated in pull request #288

ddebrunner:
was this fixed for all public api functions?

sdickes:
In my opinion we need one update -- in the Stream class getOport() should probably be renamed to _get_oport to follow the naming convention for non-public function in a public class.

PR #352, opened 2016-04-20
update getOport to _get_oport per issue #286

drhvm:
Per issue #286 changed name of getOport to _get_oport


ddebrunner:

ddebrunner:
The actual Python way is to not have getters and setters:

http://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters

http://dirtsimple.org/2004/12/python-is-not-java.html

PR #395, opened 2016-05-23
Mqttconnect

drhvm:
Support for MQTT Connector.  Provides publish and subscribe.  Currently there are still configuration parameters the require an integer vs. long that cannot be set but they are not required for basic operation.  Since the tests for this function require an external MQTT server running, I created a separate entry in build.xml to run the tests for these functions. Command "ant test.mqtt.application.api" will run the MQTT connector tests which reside in mqtest1.py in the python test directory.


ddebrunner:
Don't think you want to merge into master!

PR #396, opened 2016-05-23
Mqttconnect

drhvm:
Support for MQTT Connector. Provides publish and subscribe. Currently there are still configuration parameters the require an integer vs. long that cannot be set but they are not required for basic operation. Since the tests for this function require an external MQTT server running, I created a separate entry in build.xml to run the tests for these functions. Command "ant test.mqtt.application.api" will run the MQTT connector tests which reside in mqtest1.py in the python test directory.


ddebrunner:
What's the purpose of PyFunctionFormat ?

I think my latest changes may might it not required, you can just use transform, or at least that's the approach that probably should be taken.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,208 @@
+
+from streamsx.topology import graph
+from streamsx.topology.topology import *
+from streamsx.topology import schema
+from enum import Enum
+
+
+class mqtt_streams(object):
```
ddebrunner: Class names should normally use the CapWords convention.

https://www.python.org/dev/peps/pep-0008/#class-names

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,208 @@
...
+ #      href="http://ibmstreams.github.io/streamsx.messaging/">com.ibm.streamsx.messaging</a>
+
+
+    def __init__(self,topology,config) :
+        self.topology = topology
+        self.config = config.copy()
+        self.opCnt = 0
+        
+    def publish(self,src_op,topic=None) :
+        if topic is None:
```
ddebrunner: Why allow topic to be optional if an exception is thrown if that is the case?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,208 @@
...
+            # each op requires its own clientID
+            clientId =  params['clientID']
+            if (clientId is None and len(clientId) > 0) :
+                params['clientID'] = opCnt+"-"+clientId
+        # convert src_op outputport schema from spl po to spl rstring type
+        forOp = self.topology.graph.addOperator("com.ibm.streamsx.topology.functional.python::PyFunctionFormat")
+        forOp.addInputPort(outputPort=src_op.oport)
+        forOport = forOp.addOutputPort(schema=schema.StreamSchema("tuple<rstring message>"))                
+        op = self.topology.graph.addOperator(kind="com.ibm.streamsx.messaging.mqtt::MQTTSink")
+        oport = op.addOutputPort(schema=schema.StreamSchema("tuple<rstring message>"))
```
ddebrunner: Can we just use the existing String schema?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,208 @@
...
+        #parms['qos'] = int(0) (needs to be int vs long)
+        #del parms['messageQueueSize']
+        if topic is None :
+            parms['topicOutAttrName'] = "topic"
+        else :
+            parms['topic'] = topic
+        parms['dataAttributeName'] = "message"
+        if (++self.opCnt > 1) :
+            # each op requires its own clientID
+            clientId =  params['clientID']
```
ddebrunner: If there's no client id then let the operator select one.
The code here allows for unexpected clashes with other applications.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/runtime.py
```diff
@@ -18,6 +18,9 @@ def setupOperator(dir):
 
 def pickleObject(v):
     return pickle.dumps(v)
+
```
ddebrunner: Not required to add such a function, it's wrapping one static function in another static function for no benefit.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
drhvm: I think the MQTTSink operator requires an rstring attribute called message ie: tuple<rstring message>, in CommonSchema the attribute is called "string" ie:  tuple<rstring string>.  Using CommonSchema.String, I get the following error:
23 May 2016 16:32:31.364 [6806] ERROR #splapptrc,J[120],P[120],test_TopologyMQTTpublish_OP2 M[?:com.ibm.streams.operator.internal.compile.OperatorContextCheckerImpl.errorRequiredAttribute:-1]  - Required attribute message is missing from schema tuple<rstring string>.

ddebrunner: I guess this is the optional error output port, but there doesn't seem to be anything done with the output port. Maybe some comments indicating how this function behaves would be helpful.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
drhvm: I will change the code to: add the python object id to ensure uniqueness
params['clientID'] = id()+"-"+opCnt+"-"+clientId

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
drhvm: You are right.....the output port is really not needed.  However I still cant use  CommonSchema.String on the input port,  I think I need rstring message

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
drhvm: Changing to make topic required

drhvm:
The  PyFunctionFormat operator, depickles the source tuple and then checks to make sure it is an rstring.  It also checks for the types Long and Float and does the conversion to rstring for those types.  I am not sure if the conversion is necessary, it is more of a convenience.  We could just only allow the source to rstring and forget about the conversion of the simple types.  Then I may be able to use a standard operator like transform.  Thoughts?

ddebrunner:
Checking for specific types is not very Python like, I think simply adding a map/transform before the MQTTSink that has the output schema `CommonSchema.String` and the identity function will then convert any Python object to a rstring using Python's `str()`. See how the `publish` function does it.

drhvm:
Made changes to:
remove optional topic
change class to MqttStreams
removed PyFunctionFormat used transform instead
changed ClientId generation if multiple operators using same object

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyfunction.pm
```diff
@@ -26,6 +26,8 @@ sub splpy_tuplestyle{
     $pystyle = 'pickle';
  } elsif ($attr->getName() eq 'string') {
     $pystyle = 'string';
+ } elsif ($attr->getName() eq 'message') {
```
ddebrunner: I don't think we want a style specific to a toolkit's operator. THis can be changed to use the SPL dict approach being added in pr #400

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pywrapfunction.cgt
```diff
@@ -14,24 +14,30 @@
...
     // a function passed into the operator
     // or a pickled encoded class instance
     // represented as a string in parameter pyCallable
-    
-    <% if ($pyCallable) { %>
-      // argument is the serialized callable instance
-      Py_DECREF(appCallable);
-      appCallable = Py_BuildValue("s", <%=$pyCallable%>);
-    <%}%>
+    	<% if ($pyCallable) { %>
```
ddebrunner: Try to avoid whitepace changes, it makes it harder to review a change. 

Doesn't seem like there are changes here?????

com.ibm.streamsx.topology/com.ibm.streamsx.topology.python/namespace-info.spl
```diff
@@ -495,6 +496,79 @@ Only submission using `DISTRIBUTED`, `BUNDLE` and `STANDALONE` context type are
...
+ string tuples to MQTT topics, and
+ subscribing to MQTT topics and creating streams.
+ 
+* `MqttStreams(self, config)` MQTT Connector 
+
+A connector is for a specific MQTT Broker as specified in
+the configuration object config. Any number of  publish()and  subscribe()
+connections may be created from a single MqttStreams connector.
+
+* `publish(self, src_op, topic)` function on MQTT Conector
```
ddebrunner: why does a stream have a name of 'src_op' ?

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -293,6 +293,28 @@ namespace streamsx {
       return pyReturnVar;
     }
 
+    /**
+     * Convert a SPL blob into a Python Byte string 
+     */
+    static PyObject * pyBlobToBytes(SPL::blob & pyblob) {
```
ddebrunner: Use the existing functions: pyAttributeToPyObject

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,202 @@
...
+ # @see <a
+ #      href="http://ibmstreams.github.io/streamsx.messaging/">com.ibm.streamsx.messaging</a>
+
+
+    def __init__(self,topology,config) :
+        self.topology = topology
+        self.config = config.copy()
+        self.opCnt = 0
+        
+    def publish(self,src_op,topic) :
```
ddebrunner: Need pydoc strings.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,202 @@
...
+        parms = self.config.copy()
+        #parms['reconnectionBound'] = int(-1) (needs to be int vs long)
```
ddebrunner: Can these commented out lines be removed?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,202 @@
...
+        #del parms['messageQueueSize']
+        parms['topic'] = topic
+        parms['dataAttributeName'] = "message"
+        if (++self.opCnt > 1) :
+            # each op requires its own clientID
+            clientId =  parms['clientID']
+            if (clientId is not None and len(clientId) > 0) :
+                parms['clientID'] = clientId+"-"+str(id(self))+"-"+str(self.opCnt)
+        # convert src_op outputport schema from spl po to spl rstring type
+        forOp = src_op._map(streamsx.topology.functions.identity,schema.StreamSchema("tuple<rstring message>"))
```
ddebrunner: Still, why are we using rstring message here, doesn't rstring string work?

This is the input port right?

ddebrunner: This is what MQTTSInk says:

 *In the case where the schema contains only a signle attribute, the operator will assume that the attribute is the data attribute*

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pywrapfunction.cgt
drhvm: Sorry....no change

com.ibm.streamsx.topology/opt/python/include/splpy.h
drhvm: Function removed....no longer needed

drhvm:
Updated documentation using pydoc
got rid of the tuple<rstring message> schema now using CommonSchema.String and some functions related

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pywrapfunction.cgt
```diff
@@ -14,24 +14,30 @@
     PyObject * appCallable = 
       streamsx::topology::Splpy::loadFunction(<%=$pyModule%>, <%=$pyCallableName%>);
 
+	<% if ($pystyle eq 'string' or $pystyle eq 'message') { %>
```
ddebrunner: message is not longer a pystyle, right?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,193 @@
+
```
ddebrunner: I think this module can simply be mqtt.py

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pywrapfunction.cgt
ddebrunner: I don't think this is correct. If the passed in callable is a class and pystyle is string then _function will be set to the module, right?

I.e. the code does not work when pystyle is string and pyCallable is set.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/mqtt_streams.py
```diff
@@ -0,0 +1,193 @@
...
+            # each op requires its own clientID
+            clientId =  parms['clientID']
+            if (clientId is not None and len(clientId) > 0) :
+                parms['clientID'] = clientId+"-"+str(id(self))+"-"+str(self.opCnt)
+        # convert pub_stream outputport schema from spl po to spl rstring type
+        forOp = pub_stream._map(streamsx.topology.functions.identity,schema.CommonSchema.String)                
+        op = self.topology.graph.addOperator(kind="com.ibm.streamsx.messaging.mqtt::MQTTSink")
+        op.addInputPort(outputPort=forOp.oport)
+        #remove toolkitDir param since MQTTSink fails if set
+        del op.params['toolkitDir']
```
ddebrunner: Is this needed?

drhvm: I got this from the java implementation.  I assumed it was required to ensure unique cliendId for multiple operators using the same MqttStream class.

drhvm:
I restored the original pywrapfunction.cgt from python_spl.  Added the dictionary param.clear() in SPLGraph.addoperator() since the param dictionary retained the toolkitDir entry if it was previously set when called with a function that was not None.  Renamed mqtt_streams.py to mqtt.py

ddebrunner:
I think the issue must be that you are reusing params across multiple operators including functional and non-functional ones. There should be no need to clear in add operator.

drhvm:
I guess I don't see how that could be.  I never call addoperator passing in a params object.  It seems to me on subsequent calls to addoperator, if no param is passed in, it (params) should be empty.  However, if I put prints around the check in addoperator to see if function is not None, I see the case where the params is not empty if previously called with a function (all calls are without passing in a params object.)
code in: addoperator

```
print ("HERE",function)
if function is not None:
    params['toolkitDir'] = streamsx.topology.param.toolkit_dir()
print ("PARMS",params)
```

etc.
output looks like:
```
     [exec] HERE <function mqtt_publish at 0x7ff3f0d698c8>
     [exec] PARMS {'toolkitDir': <streamsx.topology.param.OpParam object at 0x7ff3eed3a320>}
     [exec] HERE <function identity at 0x7ff3eed321e0>
     [exec] PARMS {'toolkitDir': <streamsx.topology.param.OpParam object at 0x7ff3eed3a198>}
     [exec] HERE None
     [exec] PARMS {'toolkitDir': <streamsx.topology.param.OpParam object at 0x7ff3eed3a198>}
```

Notice the params dict retains the value set when previously called with the identity function set.  On both calls no params object was passed into the addoperator 
Im confused.

wmarshall484:
I just learned this today, but when you specify a function such as:
```python
def f(a = {}):
    print(a)
    a['hello'] = 1
```
The state of the default 'a' value is persisted across invocations of f():
```python
>>> f()
{}
>>> f()
{'hello': 1}
```

The proper way of doing this is to initialize the dictionary to 'None' and then set it if it's none:
```python
def f(a = None):
    if a is None:
        a = {}
    print(a)
    a['hello'] = 1
```
```python
>>> f()
{}
>>> f()
{}
```

ddebrunner:
Nice find Will, after a search I found this:

http://docs.python-guide.org/en/latest/writing/gotchas/

drhvm:
Updated SPLGraph.addOperator per Wills suggestion.  Thanks Will!

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionSource/PyFunctionSource_cpp.cgt
```diff
@@ -11,6 +11,10 @@
 
 <%SPL::CodeGen::implementationPrologue($model);%>
 
+<%
+  # hard code tuple stype is being used since $pystyle needed in pywrapfunction.cgt  
+  my $pystyle = 'pickle';
```
ddebrunner: This is a little confusing because pystyle refers to the style of an input port and this is a source operator with no input port.

What changed to make this required?

drhvm: Sorry, this is no longer needed.  Latest commit removes it.

PR #400, opened 2016-05-25
Spltuplesub

sdickes:
Initial changes to support subscribe from an SPL application using an spl schema.

testing has been done manually

PyFunctionHashAddr_cpp.cgt needs further investigation.

### DIFF
--- a/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
+++ b/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
@@ -95,18 +95,12 @@ void MY_OPERATOR::process(Tuple const & tuple, uint32_t port)
 print splpy_inputtuple2value($pystyle);
 %>
 
-<%if ($pystyle eq 'spltuple') {%>
+<%if ($pystyle eq 'spltupleDict') {%>
 @include "../pyspltuple2dict.cgt"
-  if (streamsx::topology::Splpy::pyTupleFilter(function_, pyDict)) {
-      submit(tuple, 0);
-  }
-  Py_DECREF(pyDict);
 <%}%>
-<%if ($pystyle ne 'spltuple') {%>
   if (streamsx::topology::Splpy::pyTupleFilter(function_, value)) {
       submit(tuple, 0);
   }
-<%}%>
 }
 
 // Punctuation processing
--- a/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
+++ b/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
@@ -97,7 +97,7 @@ print splpy_inputtuple2value($pystyle);
 
 <%if ($pystyle eq 'spltuple') {%>
 @include "../pyspltuple2dict.cgt"
-  if (streamsx::topology::Splpy::_pyTupleFilter(function_, pyDict)) {
+  if (streamsx::topology::Splpy::pyTupleFilter(function_, pyDict)) {
       submit(tuple, 0);
   }
   Py_DECREF(pyDict);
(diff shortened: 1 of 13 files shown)


com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
```diff
@@ -92,9 +95,18 @@ void MY_OPERATOR::process(Tuple const & tuple, uint32_t port)
 print splpy_inputtuple2value($pystyle);
 %>
 
+<%if ($pystyle eq 'spltuple') {%>
+@include "../pyspltuple2dict.cgt"
+  if (streamsx::topology::Splpy::pyTupleFilter(function_, pyDict)) {
+      submit(tuple, 0);
+  }
+  Py_DECREF(pyDict);
```
ddebrunner: This looks wrong, as it implies that we are holding the Python lock across the submit.

Also anything that deviates from the existing pattern makes me question why the approach is different.

ddebrunner: Isn't pyDict passed into pyTupleFunc that states it steals the reference to the argument?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyfunction.pm
```diff
@@ -22,18 +22,23 @@ sub splpy_tuplestyle{
...
     $pystyle = 'pickle';
- } elsif ($attr->getName() eq 'string') {
+ } elsif (($numattrs == 1) && ($attr->getName() eq 'string')) {
     $pystyle = 'string';
- } elsif ($attr->getName() eq 'jsonString') {
+ } elsif (($numattrs == 1) && ($attr->getName() eq 'jsonString')) {
     $pystyle = 'json';
  }
+ else {
+    $pystyle = 'spltuple';
```
ddebrunner: We will support multiple styles for a SPL tuple so maybe the name could reflect the type of the style? 'dict' ?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.pm
```diff
@@ -0,0 +1,243 @@
+use Switch;
+
+# This function takes the string of a value to be converted and its
+# type, and generates the corresponding code to convert the type from
+# c++ to Python. Note that the $type argument is a SPL::CodeGen type,
+# and not a string literal
+#
+sub cppToPythonPrimitiveConversion{
```
ddebrunner: These seems to be an almost copy of the existing code, is there a way we could refactor to have a single version?

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -228,6 +243,26 @@ namespace streamsx {
     }
 
     /*
+    * Call a function passing a Python object
+    * and return the function return as a boolean
+    * nb. the GILLock must be held
+    */
+    static int pyTupleFilter(PyObject * function, PyObject * arg) {
```
ddebrunner: The existing pyTupleFilter can be modified to just call this function, rather than having duplicated code. Much like the older _pyTupleFilter.

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -263,6 +322,9 @@ namespace streamsx {
       PyObject * arg = pyAttributeToPyObject(splVal);
 
       // invoke python nested function that generates the int32 hash
+      // clear any indication of an error and then check later for an 
+      // error
+      PyErr_Clear();
```
ddebrunner: Was this left in by mistake?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyfunction.pm
sdickes: that seems reasonable to me.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.pm
sdickes: I considered that. Let me write an issue to clean this up.

com.ibm.streamsx.topology/opt/python/include/splpy.h
sdickes: this was put in because I discovered the PyLong_AsLong() doesn't necessarily throw an exception but just sets an error -- which can be tested for later. I print the error and let the 0 hash value get returned.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/PyFunctionFilter/PyFunctionFilter_cpp.cgt
sdickes: Good catch. Given that then the code simplifies as you mention below. concerning the pyTuple... functions. I was worried about holding the GILLock and not re-obtaining it.  Let me look into this.

Issue #401, opened 2016-05-25
remove duplication of code with pyspltuple.pm


sdickes:
There is much duplication of code between 

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.pm 

and

com.ibm.streamsx.topology/opt/python/templates/common/splpy_python.pm

This needs to be corrected.

PR #406, opened 2016-05-28
Cleanup of SPL tuples to Python tuple/dict

ddebrunner:
1. Remove string based varients of Perl conversion functions - These were added due a misunderstanding (the doc is not clear), an attribute type in the perl code gen api is always a string.


sdickes:
Do these changes cover the work for # #401 ?

ddebrunner:
@sdickes I think so.

ddebrunner:
@sdickes now most of the SPl attribute to Python code is shared.

PR #407, opened 2016-05-30
Add JobConfig object.

ddebrunner:


ddebrunner:
@sdickes I haven't looked at that yet, but that is the plan.

ddebrunner:
@sdickes of course that requires some merging, the plan is to shortly have python_spl merged to master.

PR #414, opened 2016-06-13
[WIP]Add view generation to Python and Java.

wmarshall484:


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -95,6 +95,7 @@ def __init__(self, index, kind, function, name, params, graph):
         self.setParameters(params)
         self._addOperatorFunction(self.function)
         self.graph = graph
+        self.view_config = None
```
ddebrunner: Should this be using the view_config argument?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -129,6 +130,21 @@ def _map(self, func, schema):
         oport = op.addOutputPort(schema=schema)
         return Stream(self.topology, oport)
 
+    def view(self, buffer_time = 10.0, sample_size = 10000):
+        """
+        Defines a view on a stream. Returns a view object which can be graphed after the jobs is submitted.
+        """
+        name = ''.join(random.choice('0123456789abcdef') for x in range(16))
```
ddebrunner: Seems like there should be an option to have a named view.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -129,6 +130,21 @@ def _map(self, func, schema):
         oport = op.addOutputPort(schema=schema)
         return Stream(self.topology, oport)
 
+    def view(self, buffer_time = 10.0, sample_size = 10000):
+        """
+        Defines a view on a stream. Returns a view object which can be graphed after the jobs is submitted.
```
ddebrunner: Need to expand this more, what is a view. I don't really think the view object can be graphed, it provides access to the data exposed by the view, which then can be graphed. I.e that comment is very specific to one environment, Jupyter.

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
```diff
@@ -89,6 +90,25 @@ private static void portTypesNote(JSONObject op, StringBuilder sb) {
...
+    private void viewAnnotation(JSONObject op, StringBuilder sb){
+	JSONObject config = (JSONObject)op.get("config");
+	if(config == null)
+	    return;
+	JSONObject viewConfig = (JSONObject)config.get("viewConfig");
+	if(viewConfig == null)
+	    return;
+	String name = (String)viewConfig.get("name");
+	String port = (String)viewConfig.get("port");
+	Double bufferTime = (Double)viewConfig.get("bufferTime");
```
ddebrunner: Can you assume this is a Double? What if the app developer passes in 4?
Probably best to get it as a Number and then call doubleValue() on it.

Similar for sampleSize

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
```diff
@@ -89,6 +90,25 @@ private static void portTypesNote(JSONObject op, StringBuilder sb) {
         }
     }
 
+    private void viewAnnotation(JSONObject op, StringBuilder sb){
+	JSONObject config = (JSONObject)op.get("config");
+	if(config == null)
+	    return;
+	JSONObject viewConfig = (JSONObject)config.get("viewConfig");
```
ddebrunner: This layout is limiting to a single view on an operator, can multiple views exist on an operator, e.g. a view per port?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
wmarshall484: yes, accidentally added it twice.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
wmarshall484: The operator name isn't unique. There can be multiple views on a single port on a single operator. How about a naming scheme like <operator name><view number> ?

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
wmarshall484: Yes, this is a WIP

wmarshall484: Yes, this is one of the 'WIP' aspects of this pull request :)

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
wmarshall484: Fixed.

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
wmarshall484: Multiple views can now exist on an operator per port, although this doesn't happen because each time `stream.view()` is invoked, it creates an additional operator. This could be improved in the future to only use one operator for json-ifying the tuples of a stream, regardless of how many times `stream.view()` is invoked.

PR #416, opened 2016-06-14
support spl set type in subscribed schema

sdickes:
support spl set type when subscribing to spl schemas
updated spl schema subscribe to json publish test to convert the set to a list


test/java/src/com/ibm/streamsx/topology/test/python/SubscribeSPLDictTest.java
```diff
@@ -95,6 +98,20 @@ public void testSubscribeSPLDictMap() throws Exception {
             	}
         	}
 
+        	{
+        		Set<?> ex =spl.getSet("si32");
+            	JSONArray pya = (JSONArray) json.get("si32");
+            	assertEquals(ex.size(), pya.size());
+
```
ddebrunner: Can the contents of the set be tested by converting the list to a set and then assert equals on the sets?

com.ibm.streamsx.topology/opt/python/templates/common/splpy_python.pm
```diff
@@ -152,13 +150,18 @@ sub convertToPythonValue {
   if (SPL::CodeGen::Type::isList($type)) {
       return cppToPythonListConversion($iv, $type);
   }  
+  # If the type is a set, again, get the key and value types, then
```
ddebrunner: Comment has been copied from map, there is no key type with a set.

test/java/src/com/ibm/streamsx/topology/test/python/SubscribeSPLDictTest.java
```diff
@@ -99,19 +101,26 @@ public void testSubscribeSPLDictMap() throws Exception {
...
+            assertEquals(ex.size(), pya.size());
+
+            setAssertHelper(ex, pya);
         	}
 
         }
     }
+
+    private <T> void setAssertHelper(Set<T> s, JSONArray pya) {
+      Set<T> sorteds = new TreeSet<T> (s);
```
ddebrunner: Don't think you need the complexity of a TreeSet. Just create a set from the list and assert that the sets are equal. Something like ...

`assertTrue(ex.equals(new HashSet<Integer>(pya));`

ddebrunner: Ok - there is the issue that JSON will return longs rather than ints.

PR #418, opened 2016-06-14
Revert "support spl set type in subscribed schema"

ddebrunner:
Reverts IBMStreams/streamsx.topology#416

Breaks this test:

com.ibm.streamsx.topology.test.splpy.PythonFunctionalOperatorsTest

At least a compilation error:
SEVERE: CDISP9164E ERROR: An unknown type `set<int32>` was encountered when converting to back to cpp types.


sdickes:
pull request #424 contains set support and the corrected test.

PR #424, opened 2016-06-21
Splsets2

sdickes:
put change back in for subscribe with SPL sets 
ensure failing test was corrected

### DIFF
--- a/test/java/src/com/ibm/streamsx/topology/test/splpy/PythonFunctionalOperatorsTest.java
+++ b/test/java/src/com/ibm/streamsx/topology/test/splpy/PythonFunctionalOperatorsTest.java
@@ -34,9 +34,7 @@ import com.ibm.streamsx.topology.tester.Condition;
 import com.ibm.streamsx.topology.tester.Tester;
 
 public class PythonFunctionalOperatorsTest extends TestTopology {
-    
-    public static final StreamSchema ALL_PYTHON_TYPES_SCHEMA =
-            Type.Factory.getStreamSchema("tuple<boolean b," +
+  public static final String ALL_PYTHON_TYPES_STRING1="tuple<boolean b," +
     		  "int8 i8, int16 i16, int32 i32, int64 i64," +
     		  "uint8 u8, uint16 u16, uint32 u32, uint64 u64," +
     		  "float32 f32, float64 f64," +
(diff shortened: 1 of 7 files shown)


sdickes:
I am investigating a problem in this pull request -- I will comment when I determine what the problem is.

sdickes:
I have updated the logic and the tests pass but I am not sure of the intent of the tests that are in file test/java/src/com/ibm/streamsx/topology/test/splpy/PythonFunctionalOperatorsTest.java.

I see that in Noop_cpp.cgt that pyTuple is constructed but we set the converted value into pyValue. Later in ../../opt/.__splpy/common/py_functionReturnToTuples.cgt we decrement the reference to pyTuple and   not pyValue. 

Do we care?

ddebrunner:
@dlaboss The values assigned to pyValue are set into the tuple pointed to by pyTuple using PyTuple_SetItem which steals the reference, as indicated in the comment here:

https://github.com/IBMStreams/streamsx.topology/blob/master/com.ibm.streamsx.topology/opt/python/templates/common/splpy_python.pm#L182

So no need to decrement the reference for pyValue.

ddebrunner:
@dlaboss - ` I am not sure of the intent of the tests that are in file `

They are there to test the functionality of the Python SPL operator mechanism.

test/java/src/com/ibm/streamsx/topology/test/splpy/PythonFunctionalOperatorsTest.java
```diff
@@ -62,7 +60,13 @@
     		  "map<float64,int32> mf64i32," +
     		  "map<float64,uint32> mf64u32," +
     		  "map<float64,rstring> mf64r," +
-    		  "map<rstring,float64> mrf64>");
+    		  "map<rstring,float64> mrf64";
+
+    public static final StreamSchema ALL_PYTHON_TYPES_SCHEMA =
+            Type.Factory.getStreamSchema(ALL_PYTHON_TYPES_STRING1+">");
+
+    public static final StreamSchema ALL_PYTHON_TYPES_WITH_SETS_SCHEMA =
```
ddebrunner: FYI - there's a cleaner way to do this.

A schema can be extended using the extend() methods, so it's better to create the first schema and then use it to add the extra attribute with extend.

Issue #434, opened 2016-07-06
Python: Incorrect output for simple test code


Tallwhitebro:
I started off with some simple code:

**appAttempt1.py:** 
```python
from streamsx.topology.topology import Topology
import streamsx.topology.context
import appAttempt1_functions

def main():
    topo = Topology("appAttempt1")
    source = topo.source(appAttempt1_functions.readings)
    source.sink(print)
    streamsx.topology.context.submit("STANDALONE", topo.graph)

if __name__ == '__main__':
    main()
```
**appAttempt1_functions.py:**
```python
def readings():
	x = 0
	while x < 100:
		yield x
		x += 1
```
This ran with no errors, but outputted nothing of interest to the console:
```
INFO: Standalone application completed: return code=0
```

When I tried the same code with a larger controlling value in the while loop:

**appAttempt1_functions.py:**
```python
def readings():
	x = 0
	while x < 5000:
		yield x
		x += 1
```
The output was still not as expected:
```
0
1
2
3
4
5
.
. 
.
3494
3495
3496
3497
3498
Jul 06, 2016 4:21:31 PM com.ibm.streamsx.topology.internal.streams.InvokeStandalone$ProcessFuture get
INFO: Standalone application completed: return code=0
3499
```
(omitted 5-3494)

I tried another test to try to cheat around the problem in case of a strange buffering issue:

**appAttempt1_functions.py:**
```python
def readings():
	x = 0
	y = 0
	while y < 5000:
		yield y
		y+=1
	while x < 100:
		yield x
		x += 1
```
This output strangely ignores the second while loop and outputs the same as for the previous test case:
```
0
1
2
3
4
5
.
. 
.
3494
3495
3496
3497
3498
Jul 06, 2016 4:21:31 PM com.ibm.streamsx.topology.internal.streams.InvokeStandalone$ProcessFuture get
INFO: Standalone application completed: return code=0
3499
```
Any suggestions would be greatly appreciated.

sdickes:
Hi, I would suggest changing your app to this -- 

def main():
    topo = Topology("appAttempt1")
    source = topo.source(appAttempt1_functions.readings)
    source.print()
    streamsx.topology.context.submit("STANDALONE", topo.graph)

the print function is a sink that 
Prints each tuple to stdout flushing after each tuple.

ddebrunner:
It probably makes sense for the operators to flush python stdout/err when they are shutdown.

PR #447, opened 2016-07-15
Flush stdout/stderr on operator shutdown

### LINKED ISSUE #441: Flush python stdout/err when a Python operator is shutdown.
See #434

ddebrunner:
Fixes #441 

See #434 also.


sdickes:
looks good to me.

PR #471, opened 2016-08-22
Avoid copies converting attribute to Python object

### LINKED ISSUE #470: Copies of pickled python blobs are made when converting tuple to Python object.
Two copies of the blob are being made that can be avoided.

ddebrunner:
* Use reference when accessing values from the SPL tuple. 
* Use a memoryview object to pass the SPL blob attribute into Python.

Fixes #470

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/include/splpy.h
+++ b/com.ibm.streamsx.topology/opt/python/include/splpy.h
@@ -82,13 +82,13 @@ namespace streamsx {
 
 
     /**
-     * Convert a SPL blob into a Python Byte string 
+     * Convert a SPL blob into a Python Memory view object.
      */
     inline PyObject * pyAttributeToPyObject(const SPL::blob & attr) {
       long int sizeb = attr.getSize();
       const unsigned char * bytes = attr.getData();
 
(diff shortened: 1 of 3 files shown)


ddebrunner:
test now

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -82,13 +82,13 @@ namespace streamsx {
...
     /**
-     * Convert a SPL blob into a Python Byte string 
+     * Convert a SPL blob into a Python Memory view object.
      */
     inline PyObject * pyAttributeToPyObject(const SPL::blob & attr) {
       long int sizeb = attr.getSize();
       const unsigned char * bytes = attr.getData();
 
-      return PyBytes_FromStringAndSize((const char *)bytes, sizeb);
+      return PyMemoryView_FromMemory((char *) bytes, sizeb, PyBUF_READ);
```
sdickes: so for 2.7 python the PyMemoryView_FromMemory() call needs to be 
PyObject* PyBuffer_FromMemory((void *)bytes, sizeb);   ?

PR #473, opened 2016-08-22
Add python 3 to the dependency list and fix several build warnings

vdogaru:


ddebrunner:

test/java/src/com/ibm/streamsx/topology/test/api/LowLatencyTest.java
```diff
@@ -90,7 +90,7 @@ public void threadedPortTest() throws Exception{
                 lowLatencyTag = (String) placement.get(JOperator.PLACEMENT_LOW_LATENCY_REGION_ID);
             String kind = (String)op.get("kind");
             JSONObject queue = (JSONObject) op.get("queue");
-            if(queue != null && (lowLatencyTag!=null || lowLatencyTag.equals(""))){
+            if(queue != null && (lowLatencyTag == null || lowLatencyTag.equals(""))){
```
ddebrunner: I assume this was changed to a code not reached warning, but not sure if the replacement makes much sense either.

ddebrunner:
test now

ddebrunner:
test now

test/java/src/com/ibm/streamsx/topology/test/api/LowLatencyTest.java
vdogaru: Replacing `equals("")` with i`sEmpty()`:

1. `if (queue != null && (lowLatencyTag == null || lowLatencyTag.isEmpty()))`
2. `if (queue != null && (lowLatencyTag != null && !lowLatencyTag.isEmpty()))`

I'm not familiar with the graph model and cannot decide: which alternative makes more sense, (1) or (2)?

PR #492, opened 2016-08-25
Support stateful SPL operators.

ddebrunner:
Allows a callable class to become a SPL operator by decorating with `@spl.map`.

Class must currently have a __init__ method that takes no arguments (apart from `self`).

Example:

```
@spl.map(attributes=spl.PassBy.position)
class AddSeq:
    "Add a sequence number as the last attribute"
    def __init__(self):
        self.seq = 0

    def __call__(self, *tuple):
        id = self.seq
        self.seq += 1
        return tuple + (id,)
```


ddebrunner:
test now

ddebrunner:
Added support @for_each as a sink operator.

ddebrunner:
test now

PR #493, opened 2016-08-25
Added missing copyright headers.

wmarshall484:


java/src/com/ibm/streamsx/topology/spl/package-info.java
```diff
@@ -1,5 +1,9 @@
 /*
```
ddebrunner: Double copyright

PR #495, opened 2016-08-25
test and build changes to allow python 2.7 and python 3 to be used as…

sdickes:
basic build and test changes that allows the choosing of the python level (2.7 or 3.5) with the STREAMS_TOPOLOGYX_PYTHON environment variable.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -8,8 +8,7 @@ import json
 import subprocess
 import threading
 import sys, traceback
-
-import shutil
+from platform import python_version
 
 #
 # Utilities
@@ -60,6 +59,7 @@ def submit(ctxtype, graph, config = None, username = None, password = None):
     pythonrealfile = os.path.basename(pythonreal)
     pythonconfig = pythondir+"/"+pythonfile+"-config"
     pythonrealconfig = os.path.realpath(pythondir+"/"+pythonrealfile+"-config")
+    pythonversion = python_version()
 
     if config is None:
         config = {}
@@ -67,6 +67,7 @@ def submit(ctxtype, graph, config = None, username = None, password = None):
     # place the fullpaths to the python binary that is running and 
     # the python-config that will used into the config
     config["pythonversion"] = {}
+    config["pythonversion"]["version"] = pythonversion
     config["pythonversion"]["binaries"] = []
     bf = {}
     bf["python"] = pythonreal
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -73,10 +73,6 @@ def submit(ctxtype, graph, config = None, username = None, password = None):
     bf["python"] = pythonreal
     bf["pythonconfig"] = pythonrealconfig
     config["pythonversion"]["binaries"].append(bf)
-    # also write the python-config path to a file that will be used 
-    # from within pyversion.sh 
-    with open('/tmp/'+os.environ['USER']+'.pythonconfig', 'w') as pf:
-      pf.write(pythonrealconfig)
 
     fj = _createFullJSON(graph, config)
     fn = _createJSONFile(fj)
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -9,6 +9,8 @@ import subprocess
 import threading
 import sys, traceback
 
+import shutil
+
 #
 # Utilities
 #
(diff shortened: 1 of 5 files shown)


ddebrunner:

ddebrunner:
Couple of questions on STREAMS_TOPOLOGYX_PYTHON
* Why that name, since the x comes after streams in the project name? Just wondering there's an existing pattern for similar variables in any of the github toolkit. Why not use say PYTHONHOME since that is what is already used for setup?
* Who is expected to set this variable when not running in ant, e.g. pyversion.sh references it but that can be run outside of ant.

ddebrunner:
Since the plan is to support 2.7, I wonder why we don't default to just `python` rather than `python3` ?

With an anaconda install based upon Python 3.5 both python and python3 point to the same executable.

sdickes:
STREAMSX_TOPOLOGY_PYTHONHOME ?

The user also needs to set this at this point outside of our test setup as I could not find a way to pick up the correct libraries otherwise.

sdickes:
I will have to think about the anaconda case.

sdickes:
You actually mean using PYTHONHOME and using s Python binary from its bin directory.

ddebrunner:
Environment variables tend to be tricky with Streams, as you can't really depend upon them at runtime. As a user might set an env var when compiling an application but they have no control over environment variables in the instance. 

Then we have an existing standard PYTHONHOME variable that defines where Python is, how does any new variable interact with that. Note this is for how a user uses it, the envvar for the testing of topology is fine, but that's not how users interact with it.

We have two cases:

* Python application api - we should already know which Python is being run, because we are running in Python.
* SPL Operators - When the operator is created (extracted) we know what Python is being used, that should be the Python used in subsequent SPL compilations and executions.

Do we really need a new environment variable?

ddebrunner:
A minor thing on the testing setup, currently the test is tailored using ant properties of the form:
* `topology.toolkit.release`
* `topology.install.compile`

rather than environment variables or properties beginning with `streams`.

ddebrunner:
Maybe instead of changing `pyversion.sh` there should be a 2 version and a 3 version and at least for the python callables as SPL operators the operator model should be modified to call one or the other?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -50,8 +51,33 @@ def submit(ctxtype, graph, config = None, username = None, password = None):
...
+    config["pythonversion"] = {}
+    config["pythonversion"]["version"] = pythonversion
+    config["pythonversion"]["binaries"] = []
+    bf = {}
+    bf["python"] = pythonreal
+    bf["pythonconfig"] = pythonrealconfig
+    config["pythonversion"]["binaries"].append(bf)
+    # also write the python-config path to a file that will be used 
+    # from within pyversion.sh 
+    with open('/tmp/'+os.environ['USER']+'.pythonconfig', 'w') as pf:
```
ddebrunner: This seems a somewhat unsafe mechanism to pass information between entities. In a shared environment then this will break.

ddebrunner:
Is it possible to split this pull request into separate ones so we can get some of the changes merged?

There are three separate changes being made here and it seems setting the python to use during the build is ok and the dlopen part is ok and ready to merge, but there still needs to be work done on the pyversion.sh approach.

Just don't want to have to re-review changes multiple times while we head towards a solution on an indepdendent part.

sdickes:
yes I agree. I talked this over with Will and he is looking at a way to utilize the information in the JSON file but it is not there yet. I wanted to get this commited for now so I can move forward.

sdickes:
Splitting the pull or backing out the pyversion.sh change is fine -- I can let it always use python3-config for now.

Any suggestions for the pyversion.sh support needed?

ddebrunner:
Maybe we need to figure out exactly what pyversion.sh is used for, now that the library is loaded dynamically by the initialization code.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -50,8 +51,29 @@ def submit(ctxtype, graph, config = None, username = None, password = None):
     Returns:
         An output stream of bytes if submitting with JUPYTER, otherwise returns None.
     """    
+    # path to python binary
+    pythonbin = sys.executable
+    pythonreal = os.path.realpath(pythonbin)
+    pythondir = os.path.dirname(pythonbin) 
+    pythonfile = os.path.basename(pythonbin)
+    pythonrealfile = os.path.basename(pythonreal)
+    pythonconfig = pythondir+"/"+pythonfile+"-config"
```
wmarshall484: What's the purpose of `pythonconfig` compared to `pythonrealconfig`? I don't see `pythonconfig` being used.

ddebrunner:
test now

ddebrunner:
test now

ddebrunner:
I think an upgrade to Jenkins broke the github integration, I'm working on fixing it.

ddebrunner:
test now

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
sdickes: It is not needed and left over from my testing. I will remove line 60 when I do my next set of changes.

PR #506, opened 2016-08-29
Consistent region support API

ddebrunner:
Adds Java API for supporting consistent regions (starting at a source) in a topology:

* `TStream.setConsistent`
* new Java package `com.ibm.streamsx.topology.consistent`

API only, support is not yet implemented.


ddebrunner:
test now

PR #516, opened 2016-09-02
[Python] When calling submit, allow users to specify the resource url for sws manually.

wmarshall484:
If the user does *not* specify the resource URL, then an attempt is made to invoke `st geturl`. If the geturl call fails, an error is raised.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -63,7 +63,6 @@ def submit(ctxtype, graph, config = None, username = None, password = None, reso
                 process = subprocess.Popen(['streamtool', 'geturl', '--api'],
                                            stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
                 resource_url = process.stdout.readline().strip().decode('utf-8')
-                raise Exception("some exception in submit")
             except:
                 print_exception("Error getting SWS resource url ")               
                 raise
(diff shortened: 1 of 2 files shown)


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -31,7 +31,7 @@ def delete_json(fn):
 # SPL, the toolkit, the bundle and submits it to the relevant
 # environment
 #
-def submit(ctxtype, graph, config = None, username = None, password = None):
+def submit(ctxtype, graph, config = None, username = None, password = None, resource_url = None):
```
ddebrunner: Just wondering why we have specific arguments for username, password and resource_url rather than passing them as config items, since config is the configuration for the submission.

`resource_url` may be the wrong term, given that 'resource' has a specific meaning in IBM Streams (a host).

`streamtool man geturl` has this for `--api`

Specifies to print the URL for the REST API web server.

ddebrunner:
@wmarshall484 Why are these explicit parameters to submit and not config items? Just seems the submit call could get complex if every possible config item is also allowed to be a parameter to submit?

ddebrunner:
To resolve the submit parameters vs config.

The code can be clearer with the parameters, and Python supports optional parameters, thus we support an number of optional parameters on the submit. This set will grow, probably to around ten, e.g. instance, domain, etc. 

The same parameters would be accepted as config items, that's how they should be implemented, and in Java they would have to be config items (since Java does not support optional parameters as naturally as Python).

PR #519, opened 2016-09-09
Kafka build test

ddebrunner:
Add a mechanism where a test can perform a simple build check rather than complete running.

This allows a test that requires external dependencies to run to ensure that they continue to build even if the external dependencies are not set up.


ddebrunner:
test now

PR #521, opened 2016-09-12
[WIP] Add build farm support.

wmarshall484:


java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	private String getBuildFarmURL(){
+		// For the moment, just fetch a static public build farm url. 
+		// TODO: grab from VCAP file
+		return "https://streamsbuildservice.mybluemix.net/api/build/compile/spltoolkit";
+	}
+
+	private String createBuildConfig(String namespace, String mainComposite){
+		String config = "{"
```
ddebrunner: Why not create using a JSONObject ?

java/src/com/ibm/streamsx/topology/context/StreamsContext.java
```diff
@@ -49,6 +49,13 @@
          * 
          */
         ZIPPED_TOOLKIT,
+        
+        /**
+         * Execution of the topology produces the SPL application bundle
```
ddebrunner: **Streams** application bundle, not SPL.

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+		
+	}
+	
+	private File doRemoteCompile(String namespace, String mainComposite, 
+			File zippedArchive) throws ClientProtocolException, IOException {
+		String bundleName = namespace + "." + mainComposite + ".sab";
+		String archiveName = zippedArchive.getName();
+		HttpClient client = HttpClientBuilder.create().build();
+		
+		File archiveFile = new File(archiveName);
```
ddebrunner: Isn't archiveFile the same as zippedArchive?

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+					  + "}";
+		return config;
+	}
+	
+	
+    private File doPostAndRetreiveFile(CloseableHttpClient httpClient,
+            HttpRequestBase request, String bundleName) 
+            		throws ClientProtocolException, IOException {
+        request.addHeader("accept",
+                ContentType.APPLICATION_JSON.getMimeType());
```
ddebrunner: Doesn't seem like what is returned is application/json, since it seems to be the bundle?

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+
+        File bundle;
+        CloseableHttpResponse response = AnalyticsServiceStreamsContext
+        		.executeAndHandleErrors(httpClient, request);
+        HttpEntity entity = response.getEntity();
+        
+        try { 	
+            InputStream is = null;
+            FileOutputStream fos = null;
+            byte[] buffer = new byte[1024];
```
ddebrunner: Sure we can have a bigger buffer than 1k.

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+
+            try{
```
ddebrunner: A try resources statement would be simpler.

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+            	bundle = new File(bundleName);
+            	is = entity.getContent();
+            	fos = new FileOutputStream(bundle);
+
+            	// Write the bundle in chunks.
+            	for(int length; (length = is.read(buffer)) > 0;){
```
ddebrunner: read returning -1 is the definition of end-of-file, right? not zero?

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
```diff
@@ -0,0 +1,145 @@
...
+		HttpClient client = HttpClientBuilder.create().build();
+		
+		File archiveFile = new File(archiveName);
+		HttpPost postJobWithConfig = new HttpPost(getBuildFarmURL());
+        postJobWithConfig.addHeader("accept", ContentType.APPLICATION_JSON.getMimeType());
+
+        FileBody bundleBody = new FileBody(archiveFile, 
+        		ContentType.APPLICATION_OCTET_STREAM, archiveName);
+        
+        StringBody configBody = new StringBody(createBuildConfig(namespace, mainComposite),
```
wmarshall484: application/json was the first MIME type I tried, but the text/default was the only setting which worked successfully.

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
wmarshall484: Yes, it returns an `application/octet stream` response, which I process, but I'm following the template which was used in the original curl request:
```
curl -X POST \
 -H "Accept: application/json" \
 -F "inputfile=@tk8915291552598792773.zip" \
 -F 'buildconfig={
     "namespace":"myTop",
     "mainComposite":"myTop",
     "streamsVersion":"4.2",
     "pythonVersion":"3.5",
     "timeout":100
     }' \
 -v -k -O -J \
  https://streamsbuildservice.mybluemix.net/api/build/compile/spltoolkit
```

Not sure how this doesn't throw an error, since we're receiving a different MIME type than the one we specified in the Accept header.

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
wmarshall484: What size would you recommend? I picked it somewhat arbitrarily. The sab files we're retrieving can be anywhere from 1-300 MB (largest I've heard of).

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
wmarshall484: Since it's a blocking call, read will never return zero, so technically this works, but I agree
```Java 
for(int length; (length = is.read(buffer)) != -1;){
```
is more elegant.

ddebrunner: Read can in theory return 0.

ddebrunner:
test now

java/src/com/ibm/streamsx/topology/internal/context/BuildServiceCompile.java
wmarshall484: Could you elaborate on 'try resources'? Not sure I understand.

ddebrunner:
test now

PR #524, opened 2016-09-12
WIP update toolkit pyversion.sh to use correct python-config

sdickes:
WIP update to allow different python levels in the toolkit usage of pyversion.sh.

My approach for this commit is the following -- 

update json information to include an "include" for a new file pygetpythonconfig.sh
during toolkit build update pygetpythonconfig.sh to contain python version and path information,if present
pyversion.sh updated to to use pygetpythonconfig.sh to get python-config full path


ddebrunner:

ddebrunner:
test now

ddebrunner:
Is this just for the Python functions/classes as SPL operators?

Just not seeing how it works for the topology api.

sdickes:
I think it is for just the opposite -- topology api only at this point.

ddebrunner:
Could you provide some explanation how how it works, where pyversion.sh lives, where does the new modified  file pygetpythonconfig.sh live, .i.e. which toolkit?

Just not seeing how an operator in the topology toolkit calls a script in a generated application toolkit, but maybe I'm misunderstanding the scheme.

ddebrunner:
I think I understand it, but I think the code makes the assumption that the compilation will occur in the application toolkit (the $PWD in pyversion.sh) , which will not always be true. I.e. there's no requirement for the SPL compilation to be in the root of the application directory. Some of the distributed tests will probably fail with this as they create a python toolkit and then invoke its composite it from another application.

ddebrunner:
Problem,: We want to have pyversion.sh call python-config with Python 2.7 and python3-config for 3.5.

I think the only way to do this is to have Python 2 and Python 3 versions of the operators that call a different pyversion.sh. I think this can be achieved fairly easily by having identical versions of the operators in two namespaces:
`com.ibm.streamsx.topology.functional.python`  (Python 3)
`com.ibm.streamsx.topology.functional.python2` (Python 2)

The operator models are changed to use pyversion.sh from the parent directory (`../pyversion.sh`).
Then in the Python 3 namespace is a pyversion.sh that calls python3-config, and in the Python 2 namespace one that calls python-config.

The operators are identical thus the Python 2 versions can be created by copying the Python 3 version during the topology toolkit build process, an ant copy task that does the equivalent of (something like):
```
cd com.ibm.streamsx.topology
# need to exclude copying the pyversion.sh
cp -r com.ibm.streamsx.topology.functional.python/* com.ibm.streamsx.topology.functional.python2
```

Then when creating the topology the Python code uses invokes the operators it currently does, but selecting the namespace based upon the Python version running.

sdickes:
okay, let me give that a try.

PR #526, opened 2016-09-13
[WIP] adding FunctionParDo

mrshenli:
We are gradually adding operators to support (hopefully) all beam semantics. This pull request contains the counterpart for Beam's [ParDo](https://github.com/apache/incubator-beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ParDo.java) transform.

Hi @ddebrunner @wmarshall484  I am creating this pull request for code review to make sure that the design and implementation are not violating any of the existing principles. When we are all comfortable with the design, more operators will be added. Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 __pycache__
 *.sab
+*.class
+.DS_Store
+*.DS_Store
 toolkit.xml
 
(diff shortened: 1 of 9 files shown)


ddebrunner:

PR #534, opened 2016-09-20
Python2.api

sdickes:
Initial changes to have the python topology api correctly operate with python2.

This pr requires these packages installed --

future : for both python3 and python2
enum34 : for python2

and the test cluster in Poughkeepsie does not yet have these packages installed but I am working to get this updated 

The pr has changes to 
- import future to allow correct python2 operation, mostly where print() function is used. 
- c++ changes in splpy.h for method differences between python2 and 3.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/include/splpy.h
+++ b/com.ibm.streamsx.topology/opt/python/include/splpy.h
@@ -10,7 +10,6 @@
  */
 
 #define Py_LIMITED_API 0x03050000
-#include "patchlevel.h"
 
 #include "Python.h"
 #include <string>
@@ -30,9 +29,9 @@
  */
 
 #if PY_MAJOR_VERSION == 3
-#define DECL_PYTHON_LIBNAME std::string pyLib("libpython3.5m.so");
+#define TOPOLOGY_PYTHON_LIBNAME "libpython3.5m.so"
 #else
-#define DECL_PYTHON_LIBNAME std::string pyLib("libpython2.7.so");
+#define TOPOLOGY_PYTHON_LIBNAME "libpython2.7.so"
 #endif
     
 #define GET_PYTHON_VALUE_THROWIFERROR(pbytes)                            \
@@ -214,7 +213,7 @@ namespace streamsx {
         }
 
         // declare pylib and its value  
-        DECL_PYTHON_LIBNAME
+        std::string pyLib(TOPOLOGY_PYTHON_LIBNAME);
         char * pyHome = getenv("PYTHONHOME");
         if (pyHome != NULL) {
             std::string wk(pyHome);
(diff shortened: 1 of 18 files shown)


ddebrunner:

ddebrunner:
test now

sdickes:
I assume the build failed because of missing future package for python3

as noted earlier, this pr requires these packages installed --

future : for both python3 and python2
enum34 : for python2

ddebrunner:
I'll have to check since I believe the test system is using anaconda and anaconda includes future.

ddebrunner:
I think I'm mistaken on future being included in Anaconda by default, it seems it is a supported package but not in the installer, but I will verify.

ddebrunner:
@sdickes Can you explain why the pr depends on enum34 but I don't see any enum related changes?

ddebrunner:
test now

ddebrunner:
What's WIP about this pr, it seems that it's a self-contained piece of work?

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -28,7 +29,51 @@
  * Functionality for executing Python within IBM Streams.
  */
 
+#if PY_MAJOR_VERSION == 3
+#define DECL_PYTHON_LIBNAME std::string pyLib("libpython3.5m.so");
```
ddebrunner: Minor point, but this approach leaves the code non-intuitive. Since it's just `DECL_PYTHON_LIBNAME ` so it's not obvious it's defining the variable pyLib.

Seems it would be simpler to just have the #define set the name of the library, and then have the actual code be:

`std::string pyLib(TOPOLOGY_PYTHON_LIBNAME);`

com.ibm.streamsx.topology/opt/python/include/splpy.h
```diff
@@ -10,6 +10,7 @@
  */
 
 #define Py_LIMITED_API 0x03050000
+#include "patchlevel.h"
```
ddebrunner: What's this needed for? Is it because Python2 doesn't include it by default?

com.ibm.streamsx.topology/opt/python/include/splpy.h
sdickes: yes, that makes more sense. I will update.

com.ibm.streamsx.topology/opt/python/include/splpy.h
sdickes: I just tested and see it is already included -- I will remove.

sdickes:
The enum34 package installed in python2.7 allows enum usage at that level of python. 
So functions written for python3 using enum work correctly in python2.7.

ddebrunner:
So no code changes needed for supporting enums in 2.7, only that enum34 package must be in PYTHONPATH (e.g. installed by pip?)

sdickes:
Correct re. the enum34 package.

ddebrunner:
test now

sdickes:
Test cluster in Poughkeepsie updates have started. One system updated and tested.

PR #550, opened 2016-09-27
fix java python test to use selected version of python

sdickes:
this fixes the python version used in the java distributed tests

test com.ibm.streamsx.topology.test.python.SubscribeSPLDictTest works correctly now but the other two still fail. I will continue to investigate these -- 

com.ibm.streamsx.topology.test.python.PublishSubscribeStringPythonTest
com.ibm.streamsx.topology.test.python.PublishSubscribeJsonPythonTest


ddebrunner:

ddebrunner:
If I run tests directly from test/java how does topology.test.python get set by default?

sdickes:
../../common-build.xml

PR #552, opened 2016-09-29
Cleanup code to use iterators and remove unnecessary types

ddebrunner:
Cleanup that enables a next step in removing dependency on IBM's JSON4j during SPL generation.


ddebrunner:
Also use interface type for collections as the declared type.

ddebrunner:
test now

ddebrunner:
re-test as manual tests work.

PR #554, opened 2016-09-30
[WIP] Rewrite generator to use Gson and not IBM JSON4J

ddebrunner:
Trying this incrementally to ensure nothing becomes broken.

Currently portions of code are rewritten to use Gson  (`JsonObject`) while the graph is maintained as a whole with JSON4J objects `JSONObject`.

Converted code creates a temp Gson representation of the graph and works against that.


ddebrunner:
Replaced by pr #556

PR #556, opened 2016-10-04
[WIP] Use Gson for SPL code generation

ddebrunner:
1. Use gson for generation of SPL values or expression. Gson is different to JSON4J in that the original type of the object (Long, Integer etc.) is lost, so the SPL type must be entered as meta-data when created (based upon the object type).


ddebrunner:
test now

ddebrunner:
test now

ddebrunner:
test now

ddebrunner:
Going to merge this intermediate step as all the tests pass including distributed.

PR #562, opened 2016-10-10
Handle strings for Python 2

ddebrunner:


ddebrunner:
Looking more closely at the function definitions this needs some more work.

ddebrunner:
Latest commit ensures that the Python object's reference count is decremented after its byte value is copied to the rstring.

ddebrunner:
With this set of tests the Python Application API tests pass with Python 2.

There are still issues with SPL Python operators, which I'll enter issues for.

PR #564, opened 2016-10-10
[WIP] Full SPL generation using GSON

ddebrunner:
Cleanup still needed.


ddebrunner:
tested with distributed as well.

PR #565, opened 2016-10-11
decorated Python operators using python 2.7

sdickes:
for decorated python operator support use -- 
  funcsigs package for Python 2
  inspect package for Python 3 
-- when inspecting parameters 

note that com.ibm.streamsx.topology/opt/python/packages/streamsx/__init__.py was created to allow python 2.7 to be used.

### DIFF
--- a/samples/python/build.xml
+++ b/samples/python/build.xml
@@ -8,11 +8,12 @@
   <property name="streams.install" value="${env.STREAMS_INSTALL}"/>
   <property name="tk" location="../../com.ibm.streamsx.topology"/>
   <property name="pytk" location="com.ibm.streamsx.topology.pysamples"/>
+  <property name="topology.test.python" value="python3"/>
 
   <target name="all" depends="spldoc,extract"/>
 
   <target name="extract">
-   <exec executable="python3">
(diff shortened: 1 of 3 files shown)


ddebrunner:

ddebrunner:
Adding `opt/python/packages/streamsx/init.py` breaks other code.

I'll add the issue when I find it.

samples/python/build.xml
```diff
@@ -8,11 +8,12 @@
   <property name="streams.install" value="${env.STREAMS_INSTALL}"/>
   <property name="tk" location="../../com.ibm.streamsx.topology"/>
   <property name="pytk" location="com.ibm.streamsx.topology.pysamples"/>
+  <property name="topology.test.python" value="python3"/>
 
   <target name="all" depends="spldoc,extract"/>
 
   <target name="extract">
-   <exec executable="python3">
+   <exec executable="${topology.test.python}">
```
ddebrunner: This is not test code so I'm not sure it should be using a test property.

sdickes:
I removed the __init__.py file and did cleanup of .pyc files (removed them) and now python2 works.

PR #567, opened 2016-10-12
Fix paths in Makefile

ddebrunner:
* Fix paths to `spl-make-toolkit` and `sc` to avoid any dependencies on PATH
* Remove '.' from the toolkit path to avoid warning.


ddebrunner:
@wmarshall484 Look ok?

PR #578, opened 2016-10-16
Allow Topology.source to accept an Iterable instance

ddebrunner:
Simplify a source by not requiring a function but instead passing an iterable instance.

E.g.

```
# Passing an iterable from itertools
topo.source(itertools.count())

# Passing a collection
topo.source(["Hello", 2", "the", "World"])
```


ddebrunner:
test now

ddebrunner:
This makes the api a bit easier to use in Jupyter by allowing passing an iterable obtained in a cell without having to declare a function in a separate module that returns it.

PR #582, opened 2016-10-17
add __init__.py for python 2

sdickes:
Created opt/python/packages/streamsx/__init__.py
with this line
   __path__ = extend_path(__path__, __name__) 

I do not think the __path__ extend line is necessary because I also tried this with an empty file. 

I tested with both anaconda2 and anaconda3 using standard tests (decorated and api) 
and a package of mine in directory opt/python/packages/streamsx

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/__init__.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/__init__.py
@@ -1,0 +1,2 @@
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
--- a/test/python/topology/test_build_archive.py
+++ b/test/python/topology/test_build_archive.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # Licensed Materials - Property of IBM
 # Copyright IBM Corp. 2016
 import unittest


ddebrunner:

ddebrunner:
test now

ddebrunner:
I know I was seeing the Python tests failing with Python 2 unless I added an `__init__.py`. I try to see if I can reproduce it.

ddebrunner:
This is the error I see after doing this:

In Python 3 environment
```
ant clean all
cd test/python
# Run the set of python tests as Python 3 (python is a Python 3.5 executable)
ant test.python2 
```

and then in a Python 2 environment
```
cd test/python
# Run the set of python tests as Python 2 (python is a Python 2.7 executable)
ant test.python2 
```
resulting in ...
```
test.application.api:
     [echo] Topology python: python
     [exec] Traceback (most recent call last):
     [exec]   File "test_build_archive.py", line 14, in <module>
     [exec]     from streamsx.topology.topology import *
     [exec] ImportError: No module named streamsx.topology.topology

```

ddebrunner:
I have a couple of pending changes before the above steps can be run.

ddebrunner:
The issue I have goes away with this patch.

PR #589, opened 2016-10-18
do not include anaconda pre-loaded packages in SAB

sdickes:
if the running python directory indicates anaconda is being used then exclude from the SAB packages/modules that reside in the anaconda installation tree

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -87,7 +87,7 @@ def submit(ctxtype, graph, config = None, username = None, password = None, rest
     # Allows a graph or topology to be passed
     graph = graph.graph
 
-    fj = _createFullJSON(graph, config, os.path.dirname(pythondir))
+    fj = _createFullJSON(graph, config)
     fn = _createJSONFile(fj)
 
     # Create connection to SWS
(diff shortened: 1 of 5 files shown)


ddebrunner:

sdickes:
this code works but is not efficient because it works on on the fully populated list. I will now explore updating the the _DependecyResolver class

ddebrunner:
Can we use system version to see if we are running anancoda, rather than making assumptions about the install location?

http://stackoverflow.com/questions/21282363/any-way-to-tell-if-users-python-environment-is-anaconda

ddebrunner:
Sorry, I wasn't clear, but the requirement is to not add packages that Anaconda installs *by default* into the bundle, not any package that is installed under Anaconda. The issue is that pip install puts packages under anaconda, so a local install might have pip installed package `abc`, but a remote install like Bluemix that only has the out of the box Anaconda would not.

The Anaconda package list is here: https://docs.continuum.io/anaconda/pkg-docs
See the last column for if a package is in the installer.

sdickes:
got it, will make the next change based on the Anaconda package list

ddebrunner:
The list of excluded packages is just that, a list of excluded packages, not a list of conda excluded packages. When running in Anaconda that list is pre-populated by a list of Anaconda pre-installed packages. 

Then can the list be maintained by the topology object, so that an application can add to the list, rather than hidden in the dependency object which is not part of the public api.

ddebrunner:
test now

ddebrunner:
I'm not sure we need a TopologyPackages class, but that can be worked upon as we figure out the api.

I was imaging something simpler, just fields in Topology, so one could do something like:

```
topology.exclude_packages.add('blah')
```

sdickes:
That was my original thought too and the api details will be worked out soon.

PR #597, opened 2016-10-25
correct samples/python/topology/simple/echo.py

sdickes:
delete import of echo_functions as echo_functions.py has been removed


ddebrunner:

PR #607, opened 2016-11-04
Clean up log messages which are trace messages

joergboe:
Can you agree that all these messages are trace messages that require a developer’s attention.Thus these messages should not be stored in the log file.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/include/splpy.h
+++ b/com.ibm.streamsx.topology/opt/python/include/splpy.h
@@ -280,9 +280,9 @@ namespace streamsx {
         // Provide info on the setting of LD_LIBRARY_PATH
         char * pyLDD = getenv("LD_LIBRARY_PATH");
         if (pyLDD != NULL) {
-	      SPLAPPLOG(L_INFO, "LD_LIBRARY_PATH=" << pyLDD, "python");
+	      SPLAPPTRC(L_INFO, "LD_LIBRARY_PATH=" << pyLDD, "python");
         } else {
-	      SPLAPPLOG(L_INFO, "LD_LIBRARY_PATH not set", "python");
+	      SPLAPPTRC(L_INFO, "LD_LIBRARY_PATH not set", "python");
         }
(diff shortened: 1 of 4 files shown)


ddebrunner:

ddebrunner:
I would say some are log messages as it's the admin that can fix them, e.g. by changing  application environment variables.

I would say these have potential to be log messages (line numbers in splpy.h):
* 302 (maybe its associated mnessage at 298) - admin can fix the instance's environment
* 337 - admin can install the missing module/package.

joergboe:
splpy.h line 302:
I doubt that a admin is able to recover from the error reported in line 302. The comment of function loadCPython states:
// If the Python runtime is being embedded in a shared library
// (as is the case with IBM Streams), there is a bug where the 
// symbols from libpython*.*.so are not resolved properly. As
// as workaround, it's necessary to manually rediscover the
// symbols by calling dlopen().

So I would vote for a trance instead of a log here.

joergboe:
splpy.h line 337:
If the error message is put into the log file. But then I do not understand why the success message in line 341 is written into the trace file.

ddebrunner:
for line 302 the admin can correct the application instance environment variables PYTHONHOME or LD_LIBRARY_PATH.

ddebrunner:
This will be painful to review if new additional changes keep being added to the pull request. Is it possible to start using separate pull requests for independent items in the future? It makes it much quicker for them to be committed. To aid in this I would strongly encourage using branches for changes rather than your master, given this project has frequent changes from multiple contributors.

Also please try to avoid making whitespace only/reformatting changes, makes it harder to review as the code has to be checked to see if anything really changed.

ddebrunner:
splpy.h line 285 is another where the admin can fix the issue.

joergboe:
I re-introduced some logs as you requested.
Line 337 goes into logs and line 341 goes into logs
I removed the further changes for globalization

ddebrunner:
test now

PR #614, opened 2016-11-10
Cleanup submit in context.py. Useability improvements.

wmarshall484:
Submit in context.py is disorganized, especially with respect to how we handle submits:
* with and without a streams install present
* whether the user retrieves data from views
* whether the VCAP info is serialized before being sent to Java.

We also need a logging framework, so users can control the level and (and potentially location) of output when submitting.

Furthermore, the documentation is incomplete, and we don't have enums representing values users are supposed to know, e.g.:
* the "topology.service.vcap" key in config (should be VCAP_INFO)
* the "topology.service.name" key in config (should be SERVICE_NAME)
* certain submission contexts such as, "REMOTE_BUILD_AND_SUBMIT"


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -158,82 +57,199 @@ def _print_process_stderr(process, fn):
...
-            stdout_thread.daemon = True
-            stdout_thread.start()                
-            process.wait()
-            return None
-        else:            
-            return process.stdout
-    except:
-        _print_exception("Error starting java subprocess for submission")
-        raise
+def submit(ctxtype, app_topology, config = None, username = None, password = None, rest_api_url = None, log_level=logging.INFO):
```
ddebrunner: Does changing the name of the argument break backwards compatibility (graph to app_topology) ?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -158,82 +57,199 @@ def _print_process_stderr(process, fn):
...
+        app_topology: a Topology object or Topology.graph object
+        
+    Returns:
+        An output stream of bytes if submitting with JUPYTER, otherwise returns None.
+    """    
+    logger.setLevel(log_level)
+    context_submitter = ContextSubmitter(app_topology, config, username, password, rest_api_url)
+    context_submitter.submit(ctxtype)
+
+class ContextSubmitter:
```
ddebrunner: Should this be `_ContextSubmitter` to be a private class?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -158,82 +57,199 @@ def _print_process_stderr(process, fn):
...
+        self.username = username
+        self.password = password
+        self.rest_api_url = rest_api_url
+
+        if self.config is None:
+            self.config = {}
+
+        # deserialize vcap config before passing it off to Java
+        if 'topology.service.vcap' in config:
+            config['topology.service.vcap'] = json.loads(config['topology.service.vcap'])
```
ddebrunner: This seems to be expecting a serialized version of JSON inside JSON. Why not just require that the VCAP is JSON?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -158,82 +57,199 @@ def _print_process_stderr(process, fn):
...
+
+        # encode the relevant python version information into the config
+        self._do_pyversion_initialization()
+
+
+    def _do_pyversion_initialization(self):
```
ddebrunner: What actually is this used for?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/logging_utils.py
```diff
@@ -0,0 +1,22 @@
+import logging
+
+# create logger with 'spam_application'
```
ddebrunner: spam_application?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
wmarshall484: Since app_topology is passed by position, not by name, I don't believe so.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
wmarshall484: Not sure, exactly. @sdickes added this initialization code as part of the Python 2.7 support.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
wmarshall484: The refactored classes are now private.

wmarshall484:
Done with the most recent round of changes. This should be close to being ready to merge. A nice artifact of these changes is that submitting using REMOTE_BUILD_AND_SUBMIT is now much easier.

Instead of 
```python
submit("REMOTE_BUILD_AND_SUBMIT", topology, config=config, password=password, username=username,rest_api_url=rest_api_url)
```
It is now simply
```
submit("REMOTE_BUILD_AND_SUBMIT", topology, config=config)
```

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
sdickes: not that I know of @wmarshall484 
maybe a merge issue

wmarshall484: So this _do_pyversion_initialization can safely be removed?

ddebrunner:
Can you provide responses to the various comments I've made? E.g. i think changes to submit have broken backwards compat.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/logging_utils.py
wmarshall484: Fixed.

wmarshall484:
I've responded to and addressed each of your comments. WRT backwards compatibility, since app_topology is passed by position, not by name, I don't believe it breaks backwards compatibility.

wmarshall484:
Although looking at it now, changing the VCAP config to being passed as JSON, rather than a string, does break backwards compatibility.

ddebrunner:
Any named parameter can be passed by name as well, I think it needs to remain `graph`.

ddebrunner:
Sorry, missed the replies since they were hidden as outdated.

ddebrunner:
> Although looking at it now, changing the VCAP config to being passed as JSON rather than a string, does break backwards compatibility.

We need to be compatible with version 1.4, which only supported passing it as a file I think. Any functionality added in 1.5 can be changed from the EA1 release.

wmarshall484:
Even though the Java API supported the STREAMING_ANALYTICS, I don't think Python API at version 1.4 did, so I think we're fine.

wmarshall484:
Fixed the graph vs. app_topology issue.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -392,3 +393,48 @@ class UnsupportedContextException(Exception):
...
+        analytics Bluemix remote build service.
+        REMOTE_BUILD_AND_SUBMIT - the application is submitted to and built on a Bluemix streaming analytics service.
+        It is then submitted as a job.
+        TOOLKIT - Execution of the topology produces a toolkit.
+    """
+    TOOLKIT = 'TOOLKIT'
+    BUILD_ARCHIVE = 'BUILD_ARCHIVE'
+    REMOTE_BUILD_AND_SUBMIT = 'REMOTE_BUILD_AND_SUBMIT'
+    BUNDLE = 'BUNDLE'
+    STANDALONE_BUNDLE = 'STANDALONG_BUNDLE'
```
ddebrunner: Type in string variant

wmarshall484: Fixed.

PR #618, opened 2016-11-11
Globalization

joergboe:
Globalization for C++ and Java introduced


ddebrunner:

ddebrunner:
test now

ddebrunner:
Is there any recommended approach for ensuring the same message identifier is not used in Java and C++?

E..g leaving comments in each file as to the identifiers used in the other language?

ddebrunner:
I'm not seeing how `TopologyResource.h` is generated.

ddebrunner:
I think additional changes are needed to support SPL operators decorated as Python, most likely this will cause the tests to fail. 

The issue is that the message files will not be in the toolkit that contains the generated operators, maybe not `TopologyResource.h` but I'm not seeing how that is generated.

I can help out in suggesting changes, but will need to understand more about how resources work in a toolkit.

java/src/com/ibm/streamsx/topology/internal/messages/messages_en_US.properties
```diff
@@ -0,0 +1,6 @@
+# NLS_MESSAGEFORMAT_ALL
+# NLS_ENCODING=ASCII
```
ddebrunner: Curious why this is ASCII while I assume the xlf file is UTF-8?

ddebrunner:

ddebrunner:
Note from the top-level you can run `ant clean all test` to run the tests.

ddebrunner:
@wmarshall484 found this that I assume explains the generation of `TopologyResources.h`

http://www.ibm.com/support/knowledgecenter/SSCRJU_4.2.0/com.ibm.streams.dev.doc/doc/creatingatoolkitforglobalization.html

ddebrunner:
I think if you copy the `TopologyResources.h` generated file into:

`com.ibm.streamsx.topology/opt/python/include` 

it should help.

Will also need to copy the resource files into the user's toolkit, but not sure where they should go. The stuff copied into the user's toolkit is put under `opt/.__splpy`. Is it possible the resource files could be placed somewhere under there and they would still be found, or do we need to put them under `impl/nl`?

joergboe:
How can I start the test by myself?

ddebrunner:
> Note from the top-level you can run `ant clean all test` to run the tests.

joergboe:
Hard to run the test there are still jars missing. Is there a pre configured environmment? 
How can I the the details from your test (http://sdsvm0002.svl.ibm.com:8080/job/pulls.streamsx.topology/417/).

ddebrunner:
Which jars are missing?

`DEVELOPMENT.md` should be correct in describing what is needed.

joergboe:
I changed some more files. Now the spl-python-extract.py copies also the resource files.
This should work now. I verified the functionality in java part and in python part.

ddebrunner:
test now

ddebrunner:
Question from earlier, is it required that the globalization files for a toolkit are put into `impl/nl` or could they be copied under `opt/.__splpy` so that we don't pollute the user's toolkit too much?

ddebrunner:
Seeing errors such as these when using decorated Python operators:
```
SEVERE: /home/debrunne/.jenkins/jobs/pulls.streamsx.topology/workspace/test/python/spl/testtkpy/opt/.__splpy/include/splpy.h:284: error: ‘TOPOLOGY_LD_LIB_PATH’ was not declared in this scope
```

ddebrunner:
It looks like the xlf file is being copied, but not any include files. Which then are also not being generated.

joergboe:
Answer to 'question from earlier': The mechanism of the globalization is 'inherited' from the streams platform:
- The TopologySplpyResource.h file (and the TopologySplpyResource.pm) file is generated from the sc command. I have no idea if it is possible to change the destination.
- Yes the TopologySplpyResource.h is recognized if it is copied into directory opt/.__splpy.
**But** there is also the TopologySplpyResource.dat file genrated. And the platform code expects this file to be on a fixed place in directory impl/nl/.
So there is no way to avoid the impl/nl directory in the users file structure. So I decided to use the 'official' way for the globalization of a streams project:
- generate the resource files for your target locales in impl/nl/ direcory
- append the resource section in the info.xml file.
The disadvantage of this approach is that the user has to adapt the info.xml.
But a simple copy of the TopologySplpyResource.dat file into impl/nl doe not work because the 'sc -C' command cleans all *.dat files regardless.

joergboe:
Btw: the ant test command does not produce this error. Which test are you performing?

joergboe:
We must also add the info.xml with the resource section into the python/spl/testtkpy directory.
The it seems to work.
My test suite still gets some exceptions during test report generation.

ddebrunner:
But that's won't work for a user's toolkit, that's just hardcoding a fix in a test toolkit.

The decorated operators allows a user to add SPL primitive operators implemented in Python to their toolkit, we need to support their toolkit working correctly.

Can we modify/create the info.xml in their toolkit?

The very least this needs to be documented as to what the user must do.

joergboe:
- I already updated the description in com.ibm.streamsx.topology/com.ibm.streamsx.topology.python/namespace-info.spl and mentioned the resource section in info.xml file.
- Probably I can enhance the spl-python-extract scrip to:
1. Copy a default info.xml file into the target project if there is no info.xml
2. Print out a text what is to do if there is a info.xml file but this file has no Toplogy-Resource section
3. Do nothing if the info.xml has already the appropriate resource section.

joergboe:
During the execution of the test unit I got the following java excepion:

unittest:

junit.report:
[junitreport] Processing /home/joergboe/git/streamsx.topology/test/java/unittests/TESTS-TestSuites.xml to /tmp/null439163635
[junitreport] Loading stylesheet jar:file:/usr/local/apache-ant-1.9.4/lib/ant-junit.jar!/org/apache/tools/ant/taskdefs/optional/junit/xsl/junit-frames.xsl

BUILD FAILED
/home/joergboe/git/streamsx.topology/build.xml:47: The following error occurred while executing this line:
/home/joergboe/git/streamsx.topology/test/java/build.xml:321: java.lang.ExceptionInInitializerError
	at java.lang.J9VMInternals.ensureError(J9VMInternals.java:137)
	at java.lang.J9VMInternals.recordInitializationFailure(J9VMInternals.java:126)
	at com.ibm.xtq.xslt.runtime.RuntimeLibrary.getCollator(Unknown Source)
	at junit_frames_Common_P0.junit_frames$0x24$testsuites$0x24$overview$0x2e$packages$0x24$match$0x2d$template$0x2d$xsltc$0x24$(<generated>)
	at junit_frames_Common_P0.junit_frames$0x24$$0x24$applyTemplates$0x24$overview$0x2e$packages(<generated>)
	at junit_frames_Common_P0.junit_frames$0x24$testsuites$0x24$$0x24$match$0x2d$template$0x2d$xsltc$0x24$$0x24$afo$0x2d$augmented$0x24$xslt1$0x24$playback$0x2d$sax$0x2d$events_content$objectless(<generated>)
	at junit_frames_Common_P0.junit_frames$0x24$$0x24$applyTemplates$0x24$afo$0x2d$augmented$0x24$xslt1$0x24$playback$0x2d$sax$0x2d$events_content$objectless(<generated>)
	at junit_frames_Common_P0.junit_frames$0x24$$0x24$applyTemplates$0x24$afo$0x2d$augmented$0x24$xslt1$0x24$playback$0x2d$sax$0x2d$events$0x24$1(<generated>)
	at junit_frames_Common_P0.junit_frames$0x24$main(<generated>)
	at junit_frames.transform(<generated>)
	at com.ibm.xtq.xslt.runtime.AbstractTranslet.transform(Unknown Source)
	at com.ibm.xtq.xslt.jaxp.TransformerImpl.transform(Unknown Source)
	at com.ibm.xtq.xslt.jaxp.TransformerImpl.transform(Unknown Source)
	at org.apache.tools.ant.taskdefs.optional.TraXLiaison.transform(TraXLiaison.java:195)
	at org.apache.tools.ant.taskdefs.XSLTProcess.process(XSLTProcess.java:876)
	at org.apache.tools.ant.taskdefs.XSLTProcess.execute(XSLTProcess.java:414)
	at org.apache.tools.ant.taskdefs.optional.junit.AggregateTransformer.transform(AggregateTransfo

ddebrunner:
test now

com.ibm.streamsx.topology/bin/spl-python-extract.py
```diff
@@ -106,13 +114,101 @@ def copyGlobalizationResources():
...
+
+_INFO_XML_TEMPLATE="""<?xml version="1.0" encoding="UTF-8"?>
+<toolkitInfoModel
+  xmlns="http://www.ibm.com/xmlns/prod/streams/spl/toolkitInfo"
+  xmlns:common="http://www.ibm.com/xmlns/prod/streams/spl/common">
+  <identity>
+    <name>__SPLPY_TOOLKIT_NAME__</name>
+    <description>Automatic generated toolkit description file.</description>
+    <version>1.0.0</version>
+    <requiredProductVersion>4.1.0.0</requiredProductVersion>
```
ddebrunner: Is there a reason to limit this to 4.1? The current toolkit supports 4.0.1 or later.

ddebrunner:
running final tests manually and will merge once they are done.

ddebrunner:
I asked in another pull request what is `.theFileNameList` since it was missing from this PR, but now it's been added.

A quick search in the Streams docs didn't find any matches.

ddebrunner:
@joergboe I will merge this, but please provide a reply to the question about `.theFileNameList` files.

Thanks!

joergboe:
The text files for the translation team are picked up from a automated script from the production stream. The file .theFileNameList is the flag for that script, that the resource is ready for translation.
I think we can remove these files after translation completion.

ZollnaPa:
I don't recommend to remove this file. You need it for every update in globalization process in feature. If you remove it, you have to keep in mind that you must create it for every update of any message or log output.
In general, I would suggest to describe the globalisation requirements in the wiki. So you can record, why you need this file.

joergboe:
I'm how the updates will be handled in the future. But if nothing has changed in a specific resource file, why should this file be flagged with .theFileNameList?
We should discuss this but not here in this thread.

PR #639, opened 2016-12-05
Packagelists

sdickes:
Remove topologypackages.py and provide instance variables in topology.py that contain the include and exclude package sets.


ddebrunner:

ddebrunner:
test now

ddebrunner:
test now

sdickes:
I see a bug in topology.py and will be committing a fix in a little while.

sdickes:
test now

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -40,8 +49,13 @@ def __init__(self, name, files=None):
           self.opnamespace = "com.ibm.streamsx.topology.functional.python2"
         else:
           raise ValueError("Python version not supported.")
-        self.dependent_packages = topologypackages.TopologyPackages()
-        self.graph = graph.SPLGraph(name, self.dependent_packages)
+        self._conda_packages_to_exclude = set(["numpy", "scipy", "matplotlib", "h5py"])
```
wmarshall484: Just a nit, but why do you need to define `_conda_packages_to_exclude` as a field of Topology? Couldn't you just set `self.exclude_packages` to `set(["numpy", "scipy", "matplotlib", "h5py"])` on line 55?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
```diff
@@ -23,22 +21,21 @@ def __init__(self, topology_packages):
         dir = os.path.dirname(os.path.abspath(my_module.__file__))
         dir = os.path.dirname(dir)
         self._streamsx_topology_dir = dir
-        self._topology_packages = topology_packages
         
-    def add_dependencies(self, module):
+    def add_dependencies(self, module, included_packages=set(), excluded_packages=set()):
```
wmarshall484: Shouldn't this be:
```Python
def add_dependencies(self, module, included_packages=None, excluded_packages=None):
    if included_packages is None:
        included_packages = set()
    if excluded_packages is None:
        excluded_packages = set()
```

Otherwise, `included_packages` and `excluded_packages` becomes state which gets preserved over calls to add_dependencies.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
```diff
@@ -54,7 +51,7 @@ def packages(self):
         """
         return tuple(self._packages.keys())   
 
-    def _include_module(self, module):
+    def _include_module(self, module, included_packages=set(), excluded_packages=set()):
```
wmarshall484: Same comment as line 25.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
```diff
@@ -64,26 +61,28 @@ def _include_module(self, module):
...
+        for exclude_package in excluded_packages:
             if exclude_package == module.__name__ or module.__name__.startswith(exclude_package + '.'):
                 return False
             
         return True
 
 
     
-    def _add_dependency(self, module):
+    def _add_dependency(self, module, included_packages=set(), excluded_packages=set()):
```
wmarshall484: Same comment as line 25.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -117,9 +131,11 @@ class Stream(object):
     series of tuples which can be operated upon to produce another stream, as in the case of Stream.map(), or 
     terminate a stream, as in the case of Stream.sink().
     """
-    def __init__(self, topology, oport):
+    def __init__(self, topology, oport, included_packages=set(), excluded_packages=set()):
```
wmarshall484: Again, I think these should be None, and then you assign the value afterwards.

wmarshall484:
Rather than modifying the constructor of Stream and the arguments of addOperator, you could just put `include_packages` and `exclude_packages` in Graph, and then reference them inside of the addOperator call.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
sdickes: no real reason; I will change this.

sdickes:
the lists were added to Topology so a user could modify the values for the entire application. So something like this -- 


  def test_TopologySourceAndSink(self):
     topo = Topology("test_TopologySourceAndSink")
     print("topo include_packages=", topo.include_packages)
     print("topo exclude_packages=", topo.exclude_packages)
     topo.include_packages.add("frank")
     topo.include_packages.add("mabel")
     topo.exclude_packages.clear()
     topo.exclude_packages.add("elmer")
     topo.exclude_packages.add("Ida")
     print("topo include_packages=", topo.include_packages)
     print("topo exclude_packages=", topo.exclude_packages)
     hw = topo.source(test_functions.hello_world)
     hw.sink(test_functions.check_hello_world)
     streamsx.topology.context.submit("STANDALONE", topo.graph)

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
sdickes: yes

ddebrunner:
right ,Topology is the public api object, not Graph.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
sdickes: I will change

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/dependency.py
sdickes: update coming

ddebrunner:
But maybe we need to understand why we need to pass these around so much. A graph should be able to find its Topology and a dependency resolver its graph/topology.

wmarshall484:
I agree that we should be able to modify the `include_packages` and `exclude_packages` at a high level. We could use getters instead of directly accessing the list:
```Python
top = Topology()
top.get_include_packages().append("my_package")
```

The getters would just return a reference to the lists inside of Graph.
```Python
class Graph:
    ...
    def get_include_packages(self):
        return self.graph.include_packages

    def get_exclude_packages(self):
        return self.graph.exclude_packages
```

This has the benefit of not needing to pass around the parameters in the Stream constructor, and addOperator.

ddebrunner:
Python style is not to have getters, use the variables directly.

Still why do we need to pass them around, if the code can access graph then it can access the topology.

ddebrunner:
http://docs.quantifiedcode.com/python-code-patterns/correctness/implementing_java-style_getters_and_setters.html

sdickes:
I am making and testing changes to use the sets from Topology and to not pass the include and excludes sets around. I will be updating the pr soon.

sdickes:
changes have been pushed

test now

ddebrunner:
test now

PR #641, opened 2016-12-07
Use templated functions rather than code generation.

ddebrunner:
For conversion of SPL list, map and set to a Python object.

Easier to understand, debug etc.

Also results in being able to support any combination of list, map set and supported primitive types, .e.g. `list<list<float64>>`.


ddebrunner:
test now

PR #647, opened 2016-12-22
Support list,map& set of arbitrary types.

ddebrunner:
SPL Python operators, support list,map & set of arbitrary types, e.g. `list<list<float64>>`, `map<rstring,list<int32>>` etc.

- Also add checking for valid SPL schemas.
- Disallow set elements and map keys that are not hashable, so only support sets of primitive types and map keys of primitive types. This is because Python requires such items to be hashable.


ddebrunner:
test now

Issue #606, opened 2016-11-02
Modify the StreamingAnalyticsService and Distributed contexts to automatically submit to a remote build service if a local streams install is not present.


wmarshall484:
Currently, in order to choose the correct context, a user must know whether he/she is submitting from a system that contains a Streams install.

What should happen is the following -- the topology toolkit will automatically detect whether there is a Streams install and perform the correct corresponding action:

**If the system has a Streams install:**

Context | Action
------------ | -------------
Distributed | The application is compiled locally and submitted to an instance via streamtool
StreamingAnalytics | The application is compiled locally, and submitted to a BlueMix instance via REST

**If the system does *not* have a Streams install:**

Context | Action
------------ | -------------
Distributed | The application is compiled on a remote build farm and submitted to a non BlueMix instance
StreamingAnalytics | The supplied VCAP file is used to submit to a build farm within BlueMix, where the application is compiled, and from which the application is submitted as a job (most likely, this will call out to the RemoteBuildAndSubmitRemoteStreamsContext)

ddebrunner:
there should also be a config option to force a remote build even if there is a local install.

ddebrunner:
For 1.5 only need to support analytics service. Supporting distributed requires adding JMX support.

wmarshall484:
Closing to split this into two separate issues.

PR #656, opened 2017-01-09
[Issue #606] Unified STREAMING_ANALYTICS context.

wmarshall484:
When submitting from Python using the STREAMING_ANALYTICS context, one of two compilation methods will occur:
A) If there is not a local Streams install, a toolkit archive will be created and submitted to BlueMix for compilation on a build farm.
B) If there is a local Streams install, the compilation will occur locally, and the compiled .sab will be sent to run on a BlueMix instance.

If a user has a local streams install yet wishes to perform a remote build, the REMOTE_BUILD_AND_SUBMIT context may be specified explicitly.

Additionally, this PR includes a fix for a multi stage BlueMix authentication issue which prevented the connections of API requests from being established unless basic authentication was enforced by manually setting the HTTP Authentication header.


java/src/com/ibm/streamsx/topology/context/remote/RemoteContext.java
```diff
@@ -16,7 +16,8 @@
     public enum Type {
         TOOLKIT,        
         BUILD_ARCHIVE,
-        REMOTE_BUILD_AND_SUBMIT
+	REMOTE_BUILD_AND_SUBMIT,
```
ddebrunner: Isn't this context being removed?

java/src/com/ibm/streamsx/topology/internal/context/AnalyticsServiceStreamsContext.java
```diff
@@ -287,4 +316,11 @@ private BigInteger submitJobToService(File bundle, Map<String, Object> config) t
             httpClient.close();
         }
     }
+
+    static String getAPIKey(String userid, String password) throws UnsupportedEncodingException{
+	String api_creds = userid + ":" + password;
+        String apiKey = "Basic " + DatatypeConverter.printBase64Binary(api_creds.getBytes("UTF-8"));
+	return apiKey;
```
ddebrunner: Use `StandardCharsets.UTF_8` rather than `"UTF-8"`.

java/src/com/ibm/streamsx/topology/internal/context/AnalyticsServiceStreamsContext.java
```diff
@@ -211,7 +239,8 @@ private BigInteger postJob(CloseableHttpClient httpClient, JSONObject credential
         
         HttpPost postJobWithConfig = new HttpPost(url);
         postJobWithConfig.addHeader("accept", ContentType.APPLICATION_JSON.getMimeType());
-
+	postJobWithConfig.addHeader("Authorization", getAPIKey(credentials.get("userid").toString(),
```
ddebrunner: There's probably a constant for `"Authorization"`

java/src/com/ibm/streamsx/topology/internal/context/AnalyticsServiceStreamsContext.java
```diff
@@ -122,18 +142,24 @@ private JSONObject getVCAPService(Map<String, Object> config) throws IOException
...
-                credentials.get("password").toString());
-        CredentialsProvider credsProvider = new BasicCredentialsProvider();
-        credsProvider.setCredentials(
-                new AuthScope(credentials.get("rest_host").toString(),  AuthScope.ANY_PORT),
-                upc);
-        CloseableHttpClient httpClient = HttpClients.custom()
-                .setDefaultCredentialsProvider(credsProvider)
-                .build();
-        return httpClient;
+	// Commented out do to authentication issues with BlueMix when using
```
ddebrunner: Why not just remove the code?

java/src/com/ibm/streamsx/topology/context/remote/RemoteContext.java
ddebrunner: I think we want to remove this context and instead have a parameter in the config that specifies to force use of a remote build. This is because this will become an option on multiple contexts, e.g. if I just want to build a bundle using the BUNDLE context.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -450,3 +453,4 @@ class ConfigParams:
     """
     VCAP_INFO = 'topology.service.vcap'
     VCAP_NAME = 'topology.service.name'
+    REMOTE_SUBMIT = 'topology.remote_submit'
```
ddebrunner: Please be consistent with the existing properties which use camel case.

Should we change this to `topology.forceRemote` so that it can be used with other contexts?

ddebrunner: Maybe `topology.forceRemoteBuild` since that is the portion that is being forced, the submit is always to a remote system.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -52,12 +52,15 @@ def submit(ctxtype, graph, config=None, username=None, password=None, log_level=
         * BUILD_ARCHIVE - Creates a Bluemix-compatible build archive.
           execution of the topology produces a build archive, which can be submitted to a streaming
           analytics Bluemix remote build service.
-        * REMOTE_BUILD_AND_SUBMIT - the application is submitted to and built on a Bluemix streaming analytics service.
-          It is then submitted as a job.
+        * ANALYTICS_SERVICE - If a local Streams install is present, the application is built locally and then submitted
+          to a BlueMix streaming analytics service. If a local Streams install is not present, the application is
```
ddebrunner: It's "Bluemix"  not "BlueMix".

java/src/com/ibm/streamsx/topology/context/remote/RemoteContext.java
wmarshall484: The StreamsContextFactory doesn't have access to the config object when generating Contexts:
```Java
public static StreamsContext<?> getStreamsContext(StreamsContext.Type type)
```
If we modify it such that it accepts the config object:
```Java
public static StreamsContext<?> getStreamsContext(StreamsContext.Type type, Map<String, Object> config)
```
It will break the backwards compatibility of getStreamsContext.

ddebrunner: Not sure I understand, I was just expecting a config property behaving like the existing ones. It's in the `Map` passed at submit time.

wmarshall484:
I refactored the code to remove `REMOTE_BUILD_AND_SUBMIT` from all public interfaces (e.g., `Type.REMOTE_BUILD_AND_SUBMIT`). As a result, `RemoteBuildAndSubmitStreamsContext` no longer qualified as a `StreamsContext`, so I changed the class name to `RemoteBuildAndSubmitter`.

I also added documentation for the `FORCE_REMOTE` ContextProperty.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -113,9 +113,10 @@ def submit(self):
         # Otherwise, use the Java version from the streams install
         else:
             jvm = os.path.join(streams_install, "java", "jre", "bin", "java")
-            if config[ConfigParams.FORCE_REMOTE] == True:
+            if self.config[ConfigParams.FORCE_REMOTE] == True:
```
ddebrunner: Is this being treated differently to other configuration parameters, such as `topology.keepArtifacts`?

And can the name be `topology.forceRemoteBuild` as that is what is being forced, a build.

java/src/com/ibm/streamsx/topology/context/ContextProperties.java
```diff
@@ -105,4 +105,20 @@
...
     String SUBMISSION_PARAMS = "topology.submissionParams";
+
+    /**
+     * Flag to be supplied to force the compilation to occur on a 
+     * Bluemix instance, if possible. Currently only be used in
+     * conjunction with the {@code ANALYTICS_SERVICE} context.
+     * <p>
+     * Its values can be {@code true} and {@code false}. The
+     * former will force remote compilation when possible, while
+     * the latter will force local compilation, if possible.
```
ddebrunner: I think only when force is `true` should it force anything. When it's not set or `false` it should have no impact.

java/src/com/ibm/streamsx/topology/internal/context/AnalyticsServiceStreamsContext.java
```diff
@@ -63,6 +65,11 @@ public Type getType() {
     public Future<BigInteger> submit(Topology app, Map<String, Object> config)
             throws Exception {
 
+	Boolean forceRemote = (Boolean)config.get(ContextProperties.FORCE_REMOTE);
+	if(forceRemote != null && forceRemote.equals(true)){
```
ddebrunner: or more simply: `if (forceRemote != null && forceRemote)`

PR #696, opened 2017-01-22
Ensure topology.application.api standalone tests fail on error

ddebrunner:
FYI @wmarshall484 

If the standalone process failed the test would still succeed.

By returning the return code from the java submission execute back to Python we can then assert the standalone succeeded with return code 0.


ddebrunner:
test now

PR #721, opened 2017-01-31
Use python2 new-style classes in context.py

locke105:
Use the python2 new-style class syntax so we can
avoid issues with calls to super() and other inheritance issues.
This has no effect on python3, as it both class declarations
are equivalent.

For verification run the following snippet on py2 and py3:

```
  from __future__ import print_function

  class MyThing(object):
      pass

  class MyThing2:
      pass

  print(issubclass(MyThing, object))
  # True on py2 and py3

  print(issubclass(MyThing2, object))
  # False on py2, True on py3
```


ddebrunner:

PR #722, opened 2017-01-31
Convert Topology to JSON on submit and then only use JSON.

ddebrunner:
Allows contexts requiring a Streams install to use remote ones where possible, removing duplicate code and having a single consistent implementation.


ddebrunner:
Tested with these targets
test (top-level)
unittest.distributed (`test/java`)
test.python2 (`test/python`)

PR #724, opened 2017-02-02
Add ViewNotFoundError as an exception.

wmarshall484:


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/errors.py
```diff
@@ -0,0 +1,2 @@
+class ViewNotFoundError(Exception):
+    """Raised when the SWS view element cannot be located."""
```
ddebrunner: What's a 'SWS view element'?

Is it just the view cannot be located, or that no such view currently exists for a job/instance?

wmarshall484: Both. When using the SWS rest API to find a view, we iterate through the list of views found at
```
https://streams-console-c6d1.ng.bluemix.net/streams/rest/instances/{instance-id}/views/
```
If we can't find a SWS view element corresponding to the view name given when this topology is built, this exception would be thrown.

wmarshall484: >Is it just the view cannot be located, or that no such view currently exists for a job/instance

From the client side, I'm not sure if there's a way to tell the difference.

If no such view exists for an instance, then we won't be able to locate it <==> If we can't locate it, then no such view exists.

ddebrunner: What is a 'SWS view element' though?

As a user of the api I just care about a *view* and does it exist. SWS is just the mechanism used it bring the info back, and I don't know what a 'view element' is.

ddebrunner:
I was getting confused, it seems as though this should be in `streamsx.rest` not topology.

Obtaining the base view object is in rest which is where the error would be found?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/errors.py
wmarshall484: >As a user of the api I just care about a view
As a user of the topology API, this is correct. But if this is moved to streamsx.utility, in context, I think "the SWS view element cannot be located" is appropriate since that's the layer at which we're interfacing with SWS REST elements.

ddebrunner: I'm not convinced, I don't believe 'SWS' is used in the docs for the REST api.

https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.0.0/com.ibm.streams.restapi.doc/doc/restapis.html

wmarshall484: If we create streamsx.rest.errors.ViewNotFoundError as:
```Python
class ViewNotFoundError(Exception):
    """Raised when the view element resource cannot be located."""
```

Removing "SWS" and adding "resource" to conform to the terminology at [the top of this page](https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.0.0/com.ibm.streams.restapi.doc/doc/restapis-resource-element.html)

wmarshall484:
Update:
* Moved ViewNotFoundError to streamsx.rest_errors
* Moved view retrieval to streamsx.StreamsContext.retrieve_view(name=...)

wmarshall484:
Closing this PR since its changes are included in [PR 275](https://github.com/IBMStreams/streamsx.topology/pull/725)

PR #741, opened 2017-02-10
Internationalzation for languages fr it ja ko pt ru zh_CN zh_TW

joergboe:


ddebrunner:

ddebrunner:
most of the files have this:
`NLS_ENCODING=UNICODE`

What does that actually mean?

UNICODE is a character set, not an encoding. An example of encoding is `UTF-8`,  `UTF-16`, or `UCS`.

https://en.wikipedia.org/wiki/Unicode

ddebrunner:
test now

joergboe:
All *.property files must have this line
The line:
NLS_ENCODING=UNICODE
is actually required. This was required from the translation team. This line influences somehow the process how the translated files are provided.
Additionally all *.property files are converted (native2ascii) from encoded format into ascii format with Unicode escapes (\uxxxx). So the 'UNICODE' seems to logical.
Other encodings like ascii or utf-8 does not produce the desired results.

PR #761, opened 2017-02-17
Translation new language de and translation verification test script 

joergboe:


ddebrunner:

ddebrunner:
What is the test testing for?

It would be nice if the test was part of the existing testing mechanism, so it was run for an appropriate target.

joergboe:
The test script is made to support the translation verification test. It generates 3 log-messages in all available languages. These messages are presented to the translation team to verify the translation.
So it is currently nor suited for automated test

ddebrunner:
@joergboe Can you add that explanation to the test. Thanks!

PR #769, opened 2017-02-22
Ability to pass submission data back to Python from Java (e.g., JobID)

wmarshall484:


java/src/com/ibm/streamsx/topology/internal/context/remote/RemoteContexts.java
```diff
@@ -0,0 +1,38 @@
...
+    
+        // Write to the file and close the file.
+        List<String> lines = new ArrayList<>();
+        JsonObject results_json = object(submission, RemoteContext.SUBMISSION_RESULTS);
+        if(results_json == null)
+            return;
+        lines.add(results_json.toString());
+        try{          
+            Files.write(Paths.get(resultsFile), lines);
+        } catch(IOException ioe){
```
ddebrunner: Don't catch and effectively ignore exceptions, they should cause the submission to fail.

wmarshall484: Would rethrowing the exception be sufficient?

wmarshall484:
```Java
catch(IOException ioe){
    ioe.printStackTrace();
    throw ioe;
}
```

ddebrunner: Don't even catch it, let it be thrown and a higher level will catch it.

ddebrunner:
In the way that `com.ibm.streamsx.topology.internal.context.remote.DeployKeys` exists can a `SubmissionResultsKeys` be created to hold all the keys (property names) used in the submission results section of the JSON.

Along with comments explaining the scheme of how it's used.

java/src/com/ibm/streamsx/topology/internal/context/remote/RemoteContexts.java
```diff
@@ -0,0 +1,38 @@
...
+     * @return The Future<> result of the submission.
+     */
+    public static void writeResultsToFile(JsonObject submission) {
+        // Get the results file location.
+        String resultsFile = jstring(submission, RemoteContext.SUBMISSION_RESULTS_FILE);
+        if(resultsFile == null)
+            return;
+    
+        // Write to the file and close the file.
+        List<String> lines = new ArrayList<>();
```
ddebrunner: Why do you need the `List`?

Just write `results_json.toString()` directly, though ensure it is written using UTF-8 encoding.

Something like.

`Files.writer(Paths.get(resultsFile), resullts_json.toString().getBytes(StandardCharsets.UTF_8))`

Otherwise if you want to use a list then `Collections.singletonList` is cleaner.

wmarshall484:
test this

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -155,6 +155,7 @@ def submit(self):
             try:
                 _file = open(self.results_file)
```
ddebrunner: Use a `with` construct?

wmarshall484: Fixed.

wmarshall484:
Good to merge?

Issue #772, opened 2017-02-24
ConfigParams.VCAP_SERVICES does not accept filename


ibmkendrick:
The following test case does not work in streamsx 0.6.1, but was working in streamsx 0.5.5

test.py:
```python
from streamsx.topology.topology import Topology
from streamsx.topology.context import ConfigParams, submit
topology = Topology("testapp")
config = {
    ConfigParams.VCAP_SERVICES: 'vcap_services.json',
    ConfigParams.SERVICE_NAME: 'Streaming-Analytics'
}

submit("ANALYTICS_SERVICE", topology, config)
```

> python test.py
```
2017-02-10 06:53:34,920 - streamsx.topology.py_submit - INFO - Generating SPL and submitting application.
Exception in thread "main" java.lang.IllegalStateException: Not a JSON Object: "vcap_services.json"
at com.google.gson.JsonElement.getAsJsonObject(JsonElement.java:90)
at com.ibm.streamsx.topology.internal.streaminganalytics.VcapServices.getVCAPServices(VcapServices.java:62)
at com.ibm.streamsx.topology.internal.streaminganalytics.VcapServices.getVCAPService(VcapServices.java:73)
at com.ibm.streamsx.topology.internal.context.remote.RemoteBuildAndSubmitRemoteContext.submit(RemoteBuildAndSubmitRemoteContext.java:28)
at com.ibm.streamsx.topology.context.remote.RemoteContextSubmit.main(RemoteContextSubmit.java:33)
```

ddebrunner:
Strange, works for me and the tests we have.

ibmkendrick:
FYI, if 'vcap_services.json' is replaced with an absolute pathname, then it would work.

PR #776, opened 2017-02-24
Clear VCAP from config after content is loaded into env var

ibmkendrick:
Remove key ConfigParams.VCAP_SERVICES from config object after the vcap content has been loaded.  When the config object is passed into the Java submitter logic, not having this key, will trigger the load from the env var VCAP_SERVICES.

Fixes IBMStreams/streamsx.topology#772


ddebrunner:

ddebrunner:
test now

PR #797, opened 2017-03-02
Fixed munging issue with view names.

wmarshall484:


ddebrunner:
Interesting you could create a StreamsConnection where declaring the app and then use it at runtime.

How does the test actually work, since the full name of the view will be _composite_name_`::view_name` where `view_name` is passed to `view()` but the check of the name is just against `view_name`?

wmarshall484:
Because the view names are retrieved via:
```python
for view in self.sc.get_views():
             yield view.name
```

which, as far as I know, returns `view_name` and not *composite_name*`::view_name`

ddebrunner:
So `view.name` returns `composite_name.view_name`  (dot not double colon - my mistake) when the annotated operator is nested within a composite. So for operators at the main composite level `view_name` is returned, for anything else it will be `something.view_name`.

PR #815, opened 2017-03-06
Make FunctionOperatorContext public for Beam runner

mrshenli:
The Read Java operator in Streams Beam runner needs to access FunctionOperatorContext class and its constructor.


ddebrunner:

ddebrunner:
What's the purpose of making this public? It would mean the class is being used outside of the toolkit and so it would have to become part of the public api, but it's in an internal package.

Why would a Beam runner need a Streams specific class? Who would be using it through the `FunctionContext` interface?

mrshenli:
Hi @ddebrunner 

The [FunctionRead](https://github.ibm.com/streams/streams.beamrunner/blob/master/src/main/java/com/ibm/streams/beamrunner/topology/internal/functional/ops/FunctionRead.java) java operator is using the FunctionOperatorContext. It the true that the FunctionRead falls in the scope of Streamsx.Topology. The reason to introduce FunctionRead instead of using FunctionSource is that, [WindowedValue](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java) does not implement Serializable. We need to pass the Coders from Beam to Streams Java operators to perform serialization/deserialization.

mrshenli:
Hi @ddebrunner should I create another PR to add [FunctionRead](https://github.ibm.com/streams/streams.beamrunner/blob/master/src/main/java/com/ibm/streams/beamrunner/topology/internal/functional/ops/FunctionRead.java) and [streams.beamrunner.topology.internal.spljava](https://github.ibm.com/streams/streams.beamrunner/tree/master/src/main/java/com/ibm/streams/beamrunner/topology/internal/spljava) to the `Topology` API?

ddebrunner:
No.

> Work needs to be done to **design** a clean api

Issue #775, opened 2017-02-24
Standalone submission returns bundleRoot for non-existent file.


ddebrunner:
RET below is the return from a standalone submission.
```
INFO: Standalone application completed: return code=0
RET {'bundleRoot': '/home/streamsadmin/pyspl/Files.Files.sab', 'return_code': 0}
```

Two issues:

- `bundleRoot` doesn't look like the 'root' of a bundle, actually I have no idea what a bundle root is. It seems it would simply be `bundle`.
- The file does not exist since `topology.keepArtifacts` was not set, so I'm not sure of the value of returning a path to a non-existent file. Maybe only provide that path if `topology.keepArtifacts` was set to `True`?

wmarshall484:
@ddebrunner, That should have been `bundlePath`, not root.

Good point, it should only be returned if the file exists. In that case, the returned JSON would be empty except for the `return_code`.

PR #818, opened 2017-03-07
Change bundleRoot-->bundlePath

ibmkendrick:
Adds 'bundlePath' to submission result only if keepArtifacts is True.
Fixes IBMStreams/streamsx.topology#775

### DIFF
--- a/java/src/com/ibm/streamsx/topology/internal/context/BundleStreamsContext.java
+++ b/java/src/com/ibm/streamsx/topology/internal/context/BundleStreamsContext.java
@@ -72,7 +72,11 @@ public class BundleStreamsContext extends ToolkitStreamsContext {
     	    createJobConfigOverlayFile(submission, deploy, bundle.get().getParentFile());
     	
     	JsonObject results = new JsonObject();
-        results.addProperty(SubmissionResultsKeys.BUNDLE_PATH, bundle.get().getAbsolutePath());
+    	
+    	// If user asked for the SAB or asked to keep the SAB explicitly
+    	if (!byBundleUser || keepArtifacts(submission))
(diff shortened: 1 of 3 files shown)


ddebrunner:

ddebrunner:
I think this will fail in certain environments as it changed the testing approach in `test1.py` from the old mechanism to the new harness. (by the addition of the setup method).

Also tests should use the assert methods from unittest and not the `assert` statement.

The `test2*` tests use the new "recommended" approach.

ddebrunner:
In the test that doesn't set `keepArtifacts` it should probably test that the `bundlePath` property being there or not matches `keepArtifacts`, as it might be possible in the future to set the keepArtifacts outside of the test (e.g. to resolve issues).

ddebrunner:
Just realized that the bundle path needs to be present in the results even if keep artifacts is False when the topology is submitted to the BUNDLE context. I'm not sure this code will do that.

ibmkendrick:
Latest code changes applied to address the comments above:
1. move original tests to `test2.py` and make use of assert methods from `unittest`
1. In the test case, check for existence of 'bundlePath' base on `keepArtifacts` settings.
1. add BUNDLE context test in `test1.py`
1. 'bundlePath' should be in submission result for BUNDLE context as well. (Even when keepArtifacts is not specified)

ddebrunner:
Thanks for the changes, can I ask for an explicit test of submitting to the bundle context and verifying the path corrected does point to a file that exists.

Can be in its own class in test2.py since it won't be using the topology tester since it does not execute the graph.

ibmkendrick:
Latest code changes applied to address the comments above:
1. Add unit test `TestBundleMethodsNew` class in `test2.py` to test BUNDLE context
1. verifies if the file reference in result['bundlePath'] exists

ddebrunner:
test now

Issue #835, opened 2017-03-09
FAIL: test_TopologySourceFn (test2.TestDistributedTopologyMethodsNew)


ddebrunner:
```
Traceback (most recent call last):
  File "/home/streamsadmin/topology/42test/streamsx.topology/test/python/topology/test2.py", line 58, in test_TopologySourceFn
    self.assertIn('bundlePath', result)
AssertionError: 'bundlePath' not found in {'jobId': '89', 'return_code': 0}
```

ddebrunner:
So today running this passes for me.
```
python3 -m unittest test2.TestTopologyMethodsNew.test_TopologySourceFn
```

Issue #836, opened 2017-03-09
ERROR: test_BundlePath (test2.TestBundleMethodsNew)


ddebrunner:
This fails when STREAMS_INSTALL is not set, which is a valid mode for running all the tests.

Test should be skipped when it cannot be run due to environment.

ERROR: test_BundlePath (test2.TestBundleMethodsNew)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/streamsadmin/topology/42test/streamsx.topology/test/python/topology/test2.py", line 23, in test_BundlePath
    result = streamsx.topology.context.submit("BUNDLE", topo)
  File "/home/streamsadmin/topology/42test/streamsx.topology/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py", line 80, in submit
    context_submitter = _SubmitContextFactory(graph, config, username, password).get_submit_context(ctxtype)
  File "/home/streamsadmin/topology/42test/streamsx.topology/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py", line 402, in get_submit_context
    raise UnsupportedContextException(ctxtype + " must be submitted when a streams install is present.")
streamsx.topology.context.UnsupportedContextException: BUNDLE must be submitted when a streams install is present.

ibmkendrick:
Added condition to skip `TestBundleMethodsNew` when Streams install is not available.

PR #846, opened 2017-03-10
SubmissionResult to reflect keepArtifacts setting

ibmkendrick:
When `keepArtifacts=True`, submission result should contain:
* `toolkitRoot` attribute
* `bundlePath` if context is BUNDLE or streams install is available
* `archivePath`, if context is BUILD_ARCHIVE or if streams install is not available.

When `keepArtifacts=False`, submission result should not contain the above attributes, unless the context requires it.

Fixes IBMStreams/streamsx.topology#835, IBMStreams/streamsx.topology#836

### DIFF
--- a/java/src/com/ibm/streamsx/topology/internal/context/remote/BuildServiceRemoteRESTWrapper.java
+++ b/java/src/com/ibm/streamsx/topology/internal/context/remote/BuildServiceRemoteRESTWrapper.java
@@ -14,6 +14,7 @@ import java.io.IOException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Random;
+import java.util.Map.Entry;
 
 import org.apache.http.HttpEntity;
 import org.apache.http.client.ClientProtocolException;
@@ -31,6 +32,7 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.ibm.streamsx.topology.context.remote.RemoteContext;
+import com.ibm.streamsx.topology.internal.gson.GsonUtilities;
 import com.ibm.streamsx.topology.internal.streaminganalytics.RestUtils;
 
 class BuildServiceRemoteRESTWrapper {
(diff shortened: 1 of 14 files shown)


ddebrunner:

ddebrunner:
test now

test/python/topology/test2.py
```diff
@@ -11,19 +11,73 @@
 import test_vers
 
 def s4():
-    return ['one', 'two', 'three', 'four']
+   return ['one', 'two', 'three', 'four']
```
ddebrunner: We try to follow PEP8 and have four spaces as the indentation.

java/src/com/ibm/streamsx/topology/internal/gson/GsonUtilities.java
```diff
@@ -180,4 +180,24 @@ public static JsonObject objectCreate(JsonObject object, String ...property) {
...
+    
+
+    /**
+     * Get a property from a JSON object.  If the property doesn't exist, create it, and assign it an empty JSON object
+     *  
+     * @param obj source JSON object
+     * @param property property to retrieve
+     * @return The associated value of the property
+     */
+    public static JsonObject getProperty (final JsonObject obj, final String property) {
```
ddebrunner: Is this method needed, couldn't the existing `objectCreate` be used?

java/src/com/ibm/streamsx/topology/internal/context/remote/BuildServiceRemoteRESTWrapper.java
```diff
@@ -103,7 +105,10 @@ void remoteBuildAndSubmit(JsonObject submission, File archive) throws ClientProt
 			JsonObject response = doSubmitJobFromBuildArtifactPut(httpclient, deploy, apiKey, artifactId);
 			
 			// Pass back to Python
-			submission.add(RemoteContext.SUBMISSION_RESULTS, response);
+			final JsonObject submissionResult = GsonUtilities.getProperty(submission, RemoteContext.SUBMISSION_RESULTS);
+            for (Entry<String, JsonElement> entry : response.entrySet()) {
```
ddebrunner: Seems like this would be a useful utility to add to GsonUtilities since you use it twice. Add all in properties in one object to another.

ddebrunner:
For the tests that create artifacts and test the submission results contain the correct values, can you also delete the artifacts at the end of the test.

PR #865, opened 2017-03-13
Make StreamsRestClient private class

ddebrunner:
No need to have this be part of the public api.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest.py
@@ -6,7 +6,7 @@ import json
 import logging
 import streamsx.st as st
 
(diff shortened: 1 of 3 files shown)


ddebrunner:
Commit: 24624e04b33793d74ee605d5accc748f3b916999

Changed some names in context to help understanding of the code:

`app_topology` --> `graph` since at this point the object being passed around is the `SPLGraph` object the `Topology` object.

`_StreamsContextFactory`  --> `_StreamsContext` since it acts as a context, not a factory which hands out contexts.

`_RemoteBuildSubmitter` to `_StreamingAnalyticsSubmitter` - It seems this context is used to always submit to Streaming Analytics service, not just when doing a remote build.

Changed docstrings for  `_StreamingAnalyticsSubmitter`  and `_DistributedSubmitter` to represent what they actually do, I believe the descriptions for both were incorrect.

Issue #862, opened 2017-03-13
Add the ability to refresh any REST object to get the latest state.


ddebrunner:
E.g. refresh a job to get its latest health.

Code should be generic.

See comments in:

https://github.com/IBMStreams/streamsx.utility/issues/38

PR #870, opened 2017-03-14
Enable SSL certifcate verification by default

ibmkendrick:
Change summary:
* Customize StreamsConnection object with requests session arguments
* topology.context.submit config object accept StreamsConnection object
* Update unittests to disable SSL certificate for local streams connection
* Modify unittests in test/python/rest to use env variables to get credentials

Fixes IBMStreams/streamsx.topology#857

PR #871, opened 2017-03-14
Add refresh logic for REST objects

ibmkendrick:
* override `__getattr__` to access JSON response attributes without modifying the `_ResourceElement`'s attribute dictionary
* override `__setattr__` to disallow setting of attributes names that are identical with JSON response property names.

Fixes IBMStreams/streamsx.topology#862

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
@@ -37,18 +37,40 @@ class _ResourceElement(object):
     """
     def __init__(self, json_rep, rest_client):
         """
-        :param json_rep: The JSON response from a REST call.
-        :param rest_client: The client used to make the REST call.
+        Args:
+            json_rep(dict): The JSON response from a REST call.
+            rest_client(_StreamsRestClient): The client used to make the REST call.
         """
-        self.rest_client=rest_client
-        for key in json_rep:
-            if key == 'self':
-                self.__dict__["rest_self"] = json_rep['self']
-            else:
-                self.__dict__[key] = json_rep[key]
+        self.rest_client = rest_client
+        self.rest_self = json_rep.get('self', None)
+        self.json_rep = json_rep
 
     def __str__(self):
         return pformat(self.__dict__)
+
+    # Override getattr to retrieve attribute from response JSON
+    def __getattr__(self, key):
+        if 'json_rep' in self.__dict__:
+            json = self.__getattribute__('json_rep')
+            if key in json:
+                return json[key]
+        # Fallback to default behaviour
+        return self.__getattribute__(key)
+
+    # Prevent setting one of the JSON attribute into the object
+    def __setattr__(self, key, value):
+        if 'json_rep' in self.__dict__:
+            json = self.__getattribute__('json_rep')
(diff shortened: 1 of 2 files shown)


ddebrunner:

ibmkendrick:
@ddebrunner, @wmarshall484 .  Figured out an alternate way to get the JSON response properties in `_ResourceElement` without adding the properties directly into the object.

It still doesn't handle REST errors (as we have discussed earlier), but it does offer the refresh() functionality without breaking the existing design.

I will add some unit tests once #869 is merged.  (As the unit test will require StreamsConnection)

ddebrunner:
test now

wmarshall484:
+1 on overriding `__getattr__` and `__setattr__`, it doesn't pollute the Python object's `__dict__`.

ddebrunner:
Would need some tests before merging this PR.

ibmkendrick:
Waiting for #870 to be merged, before adding unit test.  (As it requires the new test infrastructure)

ddebrunner:
@wmarshall484 Can you approve(?) the ode review?

wmarshall484:
Can this be merged?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -37,19 +37,41 @@ class _ResourceElement(object):
...
+        self.rest_client = rest_client
+        self.rest_self = json_rep.get('self', None)
+        self.json_rep = json_rep
 
     def __str__(self):
         return pformat(self.__dict__)
 
+    # Override getattr to retrieve attribute from response JSON
+    def __getattr__(self, key):
+        if 'json_rep' in self.__dict__:
```
wmarshall484: Ok, now I'm confused. On this line when you call `self.__dict__`, why doesn't that recurse forever and cause a stack overflow?

wmarshall484: Since `self.__dict__` will go and call `__getattr__` again.

ibmkendrick: I think `__getattr__` is only called if the attribute is not found in the object.   So `self.__dict__` will call `__getattribute__`, but not `__getattr__`

ibmkendrick:
@ddebrunner , unit test added: eb91df1

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ibmkendrick:
```
class myobj(object):
  def __init__(self):
    self.a = 'a'

  def __getattr__(self, key):
    print ("calling __getattr__ with {}".format(key))
    return super(myobj, self).__getattribute__(key)

  def __getattribute__(self, key):
    print ("calling __getattribute__ with {}".format(key))
    return super(myobj, self).__getattribute__(key)

o = myobj()
o.a
o.__dict__
o.b
```

will yield:
```
calling __getattribute__ with a
calling __getattribute__ with __dict__
calling __getattribute__ with b
calling __getattr__ with b
```

ddebrunner:
test now

PR #873, opened 2017-03-15
Change ViewThread to _ViewDataFetcher

### LINKED ISSUE #847: ViewThread has potential issues.
`ViewThread` is a thread since it extends `threading.Thread` but it is used as a callable passed to a new `threading.Thread` object. That just looks wrong.

Shouldn't ViewThread either:

- extend Thread and be the object that is started/stopped.
- or just be a callable and use a standard Thread to execute.

The second approach might be preferable, not sure of the value of extending thread when a callable will do.

@wmarshall484 Thoughts?

ddebrunner:
Changed `ViewThread` to be a `_ViewDataFetcher` that is not a sub-class of `Thread` but a simple callable that is used as a `Thread` target.

Changed `View` to only create a `_ViewDataFetcher` when it is needed through `start_data_fetch`.

Fixes #847

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
@@ -115,11 +115,11 @@ class _StreamsRestClient(object):
         return pformat(self.__dict__)
 
 
-class ViewThread(threading.Thread):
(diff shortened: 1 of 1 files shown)


ddebrunner:
@wmarshall484 I added #879 to track bring the samples over & fixing them.

PR #877, opened 2017-03-15
Remove get_views() and get_view()

### LINKED ISSUE #856: Support for getting a list of views, and getting a list of views by name.
TODO: since views all have different physical names (e.g., instancee34f9daad::job02::myView), but can have the same logical names (myView), the get_view(str) method should return a list. Currently it just finds the first view encountered.

Maybe we should support a way of finding a view based on the physical name, such as searching for a view within a particular instance or job.

ddebrunner:
At this point removing these methods as:

- `get_views()` breaks if an instance visible in the domain is not running.
- `get_view()` does not work for views that are within composites.

Rather than try to fix these at this point (close to a release) remove them and they can be added back later, especially once we have a strategy for how to accurately match view "logical" names to the name reported in the REST call. (or do we match by id?).

One can still obtain a list of views for an instance or a job. 

Fixes #856 (at least for this release).

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest.py
@@ -135,33 +135,6 @@ class StreamsConnection:
         """
         return self._get_elements('installations', Installation)
 
-    def get_views(self):
-        """Gets a list of all View resources across all known streams installations.
-
-        :return: Returns a list of all View resources.
-        :type return: list.
-        """
-        views = []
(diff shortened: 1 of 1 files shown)


wmarshall484:
This sample in streamsx.utility needs to be rewritten, but +1

https://github.com/IBMStreams/streamsx.utility/blob/master/python/samples/Rest_API_View_Sample.ipynb

ddebrunner:
@wmarshall484 I added #879 to track bring the samples over & fixing them.

PR #931, opened 2017-03-24
Reasonable name handling for Python.

ddebrunner:
With this app:
```
topo = stt.Topology()
s = topo.source(['a'], name="VehicleLocations")
s = s.filter(lambda _:True, name="IdleVehicles")
s = s.map(lambda _:_, name="BrokenDown")
s = s.for_each(lambda _:None, name="Alerts")
```

Before and after console view (upper is new)

image

and the output ports have the simpler name:

image


ddebrunner:
Work is still needed on the input ports.

PR #934, opened 2017-03-25
[WIP] Fix tweeepy app using application configuration for credentials.

### LINKED ISSUE #932: Accessing streamsx.ec can fail when using topology and 'dilled' class instances.
The class has a reference to hte module  `streamsx.ec` and dill preserves the state of the module which has `_supported = False`, as it is only supported when running inside a PE.

This not supported state is preserved through dill and thus on the Streams instance side the module becomes unsupported.

This causes a modified version of the tweepy app to fail that uses an application configuration.

ddebrunner:
Fixes #932


ddebrunner:
Replaced by pr #937

Issue #864, opened 2017-03-13
Document Python streamsx.rest api.


ddebrunner:

ibmkendrick:
I'll look at `StreamsConnection` and `StreamingAnalyticsConnection`

ddebrunner:
I'll do `topology.context` and `spl.*`

markheger:
documentation available: https://streamsxtopology.readthedocs.io/en/stable/streamsx.rest.html

PR #941, opened 2017-03-28
Add docstring. #864

ibmkendrick:


ddebrunner:

ddebrunner:
test now

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -258,58 +321,122 @@ def get_view_items(self):
 
 
 class ViewItem(_ResourceElement):
-    """
-    Represents the data of a tuple, it's type, and the time when it was collected from the stream.
+    """Represents the data of a tuple, it's type, and the time when it was collected from the stream.
```
ddebrunner: "its" type, not "it's"

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -62,21 +67,25 @@ def __setattr__(self, key, value):
         super(_ResourceElement, self).__setattr__(key, value)
 
     def refresh(self):
-        """Make a REST call to refresh the JSON object stored within the object
-        Returns:
-            _ResourceElement: The provided object instance is returned with updated value
+        """Refresh the JSON representation of the resource for the latest status
```
ddebrunner: Why is it the "JSON represresentation"?

Isn't it just refreshing the object, the fact that JSON was used is an implementation detail isn't it? By the time the user gets the object it's not JSON, just a Python object.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -93,7 +102,20 @@ def _get_elements(self, url, key, eclass, id=None, name=None):
         return elements
 
     def _get_element_by_id(self, url, key, eclass, id):
-        """Get a single element matching an id"""
+        """Get a single element matching an `id`
```
ddebrunner: These messages are not appearing in the Pythondoc output I hope, since they are _ methods.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -229,27 +265,54 @@ def __init__(self, json_view, rest_client):
...
+            :py:class:`queue.Queue`: Queue containing view data
+        """
         self.stop_data_fetch()
         self._data_fetcher = _ViewDataFetcher(self, self._tuple_fn)
         t = threading.Thread(target=self._data_fetcher)
         t.start()
         return self._data_fetcher.items
 
     def get_view_items(self):
+        """Get a list of ViewItem resources associated with this view resource
```
ddebrunner: I don't think we need 'resource' everywhere. At this point it's a ViewItem and a View.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -258,58 +321,122 @@ def get_view_items(self):
...
+            name(str, optional): only return views containing property **name** that matches `name`.  `name` can be a
+                regular expression.  If `name` is not supplied, then all view resources are returned.
+
+        Returns:
+            :py:obj:`list` of :py:class:`View`: list of view resources matching `name`.
+        """
         return self._get_elements(self.views, 'views', View, name=name)
 
     def get_domain(self):
+        """Get the Streams domain that owns this job element resource
```
ddebrunner: Get the Streams domain for the instance that this job is running in.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -258,58 +321,122 @@ def get_view_items(self):
...
     def get_operator_connections(self):
+        """Get the list of OperatorConnection element associated with this job resource
+
+        Returns:
+            :py:obj:`list` of :py:class:`OperatorConnection`: list of OperatorConnection resource objects.
+        """
         return self._get_elements(self.operatorConnections, 'operatorConnections', OperatorConnection)
 
     def get_operators(self):
+        """Get the list of Operator element associated with this job resource
```
ddebrunner: Same comment as resource but for element, at this point it's a list of operators.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -618,11 +908,11 @@ def cancel_job(self, job_id=None, job_name=None):
         """Cancel a running job.
 
         Args:
-            job_id(str): Identifier of job to be canceled.
-            job_name(str): Name of job to be canceled.
+            job_id (str, optional): Identifier of job to be canceled.
```
ddebrunner: Is this 'optional' a standard approach. I thought the fact that the method signature showed it was optional was sufficient? Just not seen it before.

ddebrunner:
Do we need the :py:class: in this case. I thought in Args and Returns Sphinx tries to figure out what the reference is and auto-inserts it, e.g. with a arg this is sufficient : `graph(Topology)`.
```
Returns:
             :py:class:`Instance`: 
```

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ibmkendrick: The JSON is an implementation detail, but it's also a necessary bit of information for user to access the properties.  When user print a resource element, this is what they see:
```
{
  'rest_self': 'https://blahblah',
  'json_rep': { 'resourceType': 'blah', 'name': 'blah', ... }
}
```

The JSON representation is exposed as the `json_rep` property.  They can access the properties via dot notation.  (i.e. element.resourceType).  I'd like to find a better description to explain this.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ibmkendrick: None of the _methods are exposed in pythondoc.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: I thought that all those fields were pushed to the top level, so they are not expected to access `json_rep`?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ibmkendrick: I got this from the google style doc: http://www.sphinx-doc.org/en/stable/ext/example_google.html  They use `optional` to indicate that an argument is optional.  (Though I agree the method signature gives the same information as well)

ibmkendrick:
Latest commit addresses the above comments:
* remove `:py:` from Args/Returns
* add attribute docstring for all resources elements
* update docstring best practices on `DEVELOPMENT.md`

ddebrunner:
test now

ddebrunner:
test now

ibmkendrick:
Ran: `python -m unittest discover`, and all tests passed.

PR #960, opened 2017-04-05
Code to handle pass by reference for Source,Map,ForEach for Python objects.

ddebrunner:
- Setup `outputConnections` count parameter for functional Python operators to indicate how many downstream functional operators the output port is connected to. Only set if connected only to functional operators.
- Handle a `PyObject *` being passed as the SPL tuple. Currently in the PR no tuples are sent this way, but the handling of input tuples was added to ensure all the tests pass (incremental development).

Next step would be to start passing tuples using `PyObject *` when `outputConnections` is positive and the port is not connected to a PE output port.


ddebrunner:
Current state is that `Source` and `Map` operators submit tuples by reference when appropriate.

To pass by reference a `Source` or `Map` must be connected to a `Map` or `ForEach` as those are currently the only ones that fully handled a Python object reference (`PyObject *`) on their input ports.

And pass-by-ref is disabled when a port is connected to a PE output port (or not connected  at all).

ddebrunner:
Make this a viable PR, suggest follow on support for Filter/FlatMap be done in separate pull requests to make reviewing easier.

com.ibm.streamsx.topology/opt/python/include/splpy_tuple.h
```diff
@@ -36,14 +48,56 @@ namespace streamsx {
 
   /**
    * Convert the SPL tuple that represents a Python object
-   * to a Python tuple that holds:
-   *  - pickled value as a memory view object
+   * to a Python tuple that holds as its first argument.
+   * One or two arguments in a Python tuple are passed to the function
+   * (which is a wrapper around the user function).
```
wmarshall484: There are some grammatical errors and missing words which make this section difficult to understand:
* `to a Python tuple that holds as its first argument.` should end in a colon
* `One or two arguments in a Python tuple are passed to the function` I think should be ` `One or two arguments in a Python tuple which are passed to the function`

com.ibm.streamsx.topology/opt/python/include/splpy_tuple.h
```diff
@@ -36,14 +48,56 @@ namespace streamsx {
...
+   * 
+   *  Python object was passed by reference in the SPL tuple
```
wmarshall484: Nit, could you start the sentence with `if`? 

I had to read this a couple times to understand that this was a description of the structure of the tuple based on whether the SPL blob contained a reference or a pickled object.

com.ibm.streamsx.topology/opt/python/include/splpy_tuple.h
```diff
@@ -36,14 +48,56 @@ namespace streamsx {
...
+          pyTuple = PyTuple_New(1);
+          PyTuple_SET_ITEM(pyTuple, 0, value);
+      }
+      else if (fmt == STREAMSX_TPP_PICKLE) {
+          PyObject * value = pySplValueToPyObject(pyo);
+
+          pyTuple = PyTuple_New(2);
+          PyTuple_SET_ITEM(pyTuple, 0, value);
+
+          // Pass a non-None value is as the "pickle marker (pm)"
```
wmarshall484: Should be Pass a non-None value `in`

ddebrunner: Should be "Pass a non-None value ~~is~~ as the ..."

java/src/com/ibm/streamsx/topology/generator/spl/Optimizer.java
```diff
@@ -0,0 +1,136 @@
...
+                        }
+                }
+
+                if (canPassByRef)
+                    connCounts[port] = conns.size();
+            }
+
+            boolean paramNeeded = false;
+            for (int i = 0; i < connCounts.length; i++) {
+                if (connCounts[0] != -1) {
```
wmarshall484: Should this be `connCounts[i]` not `connCounts[0]` ?

wmarshall484: Since we only support one output port right now, I don't think this breaks anything yet.

PR #998, opened 2017-04-28
Initial set of REST API's that are similar to the python REST api 

yvchan:
Contains most of the basic functions to get a StreamingAnalyticsConnection and a StreamsConnection.
From these connections, you can get a list of instances, and follow through the chain with a list of jobs, a list of operators and a list of metrics.  In addition, you can identify a specific instance, or job.  

What's still to be added in: 

- package-info.java
- better exception handling.  currently the code throws IllegalStateException for any httpstate != 200.  these should be improved for better granularity.
- testcases


ddebrunner:

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
```diff
@@ -0,0 +1,117 @@
...
+        JsonObject credential = new JsonObject();
+        credential = service.get("credentials").getAsJsonObject();
+
+        String userId = credential.get("userid").getAsString();
+        String authToken = credential.get("password").getAsString();
+        String resourcesPath = credential.get("resources_path").getAsString();
+        String sURL = credential.get("rest_url").getAsString() + resourcesPath;
+
+        String restURL = "" ;
+        StreamingAnalyticsConnection SAConn = new StreamingAnalyticsConnection( userId, authToken, restURL ) ;
```
ddebrunner: Java style is to use lower/camel case for variable names. 

For a typical Java developer `SAConn` looks like a class name and makes the code jarring to read.

java/src/com/ibm/streamsx/rest/primitives/InstancesArray.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.gson.Gson;
+import com.ibm.streamsx.rest.StreamsConnection;
+import com.ibm.streamsx.rest.primitives.Instance;
+import com.ibm.streamsx.rest.primitives.InstanceGson;
+
+public class InstancesArray {
```
ddebrunner: Can the purpose/role of these `*Array` classes be explained?

Somewhat of an anti-pattern to have to have such a class, would have just expected `List<Instance>`.

java/src/com/ibm/streamsx/rest/primitives/JobGson.java
```diff
@@ -0,0 +1,39 @@
+/*
+# Licensed Materials - Property of IBM
+# Copyright IBM Corp. 2017
+ */
+package com.ibm.streamsx.rest.primitives;
+
+import java.util.ArrayList;
+
+class JobGson {
```
ddebrunner: Not sure such a class is needed. I thought Gson could automatically create instances of a class from JSON.

java/src/com/ibm/streamsx/rest/primitives/InstancesArray.java
yvchan: When querying for the list of instances, the json structure that comes back includes a an array of instances, a resource type and the total number of instances coming back.   GSON needed a structure that it could understand so that I didn't need to manipulate the JSON to get to the array.   

Looking at this, I can clean up the code in the array objects some more, and make them package level objects only, not public.

java/src/com/ibm/streamsx/rest/primitives/JobGson.java
yvchan: Gson is able to, as long as the class itself is simple enough.  If it gets too complex, we start needing to add in serializers & deserializers, or by setting up annotations on what is included/not included.  By de-coupling the actual json structure from what is used by the client applications, this is a cleaner implementation, and if the structure changes, it's impact is contained.

PR #1004, opened 2017-05-02
allow a job to be canceled from the Job object in the rest api.

yvchan:
*Array* classes have been cleaned up to remove an unused StreamsConnection object, and most have been made package level classes.  The InstanceArray is needed by the higher level package, so it's been left public.

The Job.cancel() functionality has been added to the api, and can cancel either a StreamingAnalyticsConnection job (via the REST api), or a StreamsConnection job (via the InvokeCancel api)

### DIFF
--- a/java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
+++ b/java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
@@ -4,20 +4,22 @@
  */
 package com.ibm.streamsx.rest;
 
-import java.io.IOException;
-import java.util.List;
-
 import static com.ibm.streamsx.topology.context.AnalyticsServiceProperties.SERVICE_NAME;
 import static com.ibm.streamsx.topology.context.AnalyticsServiceProperties.VCAP_SERVICES;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.logging.Logger;
+
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
-import com.ibm.streamsx.rest.primitives.Instance;
-import com.ibm.streamsx.rest.primitives.Job;
-import com.ibm.streamsx.rest.primitives.Metric;
-import com.ibm.streamsx.rest.primitives.Operator;
 import com.ibm.streamsx.topology.internal.streaminganalytics.VcapServices;
 
-public class StreamingAnalyticsConnection extends StreamsConnection{
+public class StreamingAnalyticsConnection extends StreamsConnection {
+
+    static final Logger traceLog = Logger.getLogger("com.ibm.streamsx.topology.rest.StreamingAnalyticsConnection");
+
+    private String jobsPath;
 
     /**
      * Basic connection to the Streaming Analytics Instance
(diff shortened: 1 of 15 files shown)


ddebrunner:

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
```diff
@@ -29,50 +42,91 @@
...
+        jobsPath = sJobs;
+    }
+
+    /**
+     * @param jobId
+     *            string indicating the job id to be cancelled
+     * @return true if job is cancelled, false otherwise
+     */
+    @Override
+    public boolean cancel_job(String jobId) {
```
ddebrunner: Java style for methods would be `cancelJob`.

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
```diff
@@ -29,50 +42,91 @@
...
+            if (HttpStatus.SC_OK == rcResponse) {
+                sReturn = EntityUtils.toString(hResponse.getEntity());
+                rc = true;
+            } else {
+                rc = false;
+            }
+            traceLog.finest("Request: [" + deleteJob + "]");
+            traceLog.finest(rcResponse + ": " + sReturn);
+        } catch (IOException e) {
+            rc = false;
```
ddebrunner: Don't hide exceptions, it causes immense pain for support.

java/src/com/ibm/streamsx/rest/primitives/InstancesArray.java
```diff
@@ -12,13 +12,15 @@
 import com.ibm.streamsx.rest.primitives.Instance;
 import com.ibm.streamsx.rest.primitives.InstanceGson;
 
+/**
+ * class to hold information about instances from GET instances URL
+ * this class is public, but should only be used by classes in the rest package
```
ddebrunner: Can it be made package private?

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
yvchan: will change this.

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnection.java
yvchan: ok, will remove the catch, and set the cancel to throw...

java/src/com/ibm/streamsx/rest/primitives/InstancesArray.java
yvchan: as it is, the primitives are a sub-directory currently.  If I move them up a level, then yes, they can be made package private.  I've only exposed the one InstanceArray as public right now, since the Connection objects need this as the entry point.   Let me know if this is what you want, and I'll make the changes for it.

ddebrunner: Maybe just have a single package.

yvchan:
Dan, I've replaced the commits in this branch with new ones after the conversation here.

1) the classes are now all at the same level in com.ibm.streamsx.rest so that I can make certain functions package level only, and have exposed only what should be used by end users as public.
2) the cancelJob has been made package level in the connection objects so that the only way to cancel a job is via the actual job object itself.

ddebrunner:
> the cancelJob has been made package level in the connection objects so that the only way to cancel a job is via the actual job object itself.

What was the reason for this?

yvchan:
This makes it more likely the job being canceled is the one you intend to cancel, and you can't cancel just any job number...

ddebrunner:
Not sure how, all the other apis allow a job to be cancelled by just its id. Not sure why the Java api would be different.

yvchan:
ok.  i can make it public.

PR #1008, opened 2017-05-04
add a userName to the invokeCancel()

yvchan:
the current InvokeCancel call in topology/internal/streams does not take a userName, but uses the name of the current user to attempt to cancel the job.  This change will push the name that was used to create the streamsConnection object down so that we use that userid instead of the user that's running the application.

this doesn't allow for the password to be passed down, so if the password is needed, the InvokeCancel will throw an exception and fail.

### DIFF
--- a/java/src/com/ibm/streamsx/topology/internal/streams/InvokeCancel.java
+++ b/java/src/com/ibm/streamsx/topology/internal/streams/InvokeCancel.java
@@ -18,10 +18,17 @@ public class InvokeCancel {
     static final Logger trace = Topology.STREAMS_LOGGER;
 
     private final BigInteger jobId;
+    private final String userName;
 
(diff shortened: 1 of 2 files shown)


ddebrunner:

java/src/com/ibm/streamsx/topology/internal/streams/InvokeCancel.java
```diff
@@ -18,10 +18,17 @@
...
 
+    public InvokeCancel(BigInteger jobId, String userName) {
+        super();
+        this.jobId = jobId;
+        this.userName = userName;
+    }
     public InvokeCancel(BigInteger jobId) {
         super();
         this.jobId = jobId;
+        this.userName = "";
```
ddebrunner: It's usual in Java to have `null` represent an unset value, not the empty string.

PR #1014, opened 2017-05-05
add getInputPorts and getOutputPorts to Operator class

yvchan:
Changed the getInputPorts and getOutputPorts to return lists of InputPort and OutputPort respectively.

Calling the getMetrics from these ports will return a list of metrics for the port.


ddebrunner:

PR #1016, opened 2017-05-08
remove unnecessary *Gson.java files

yvchan:
remove the unnecessary *Gson.java files and use the @Expose annotation from Gson to control what is in/out of the JSON object created.

move setConnection functions into the package as the end user of this code should not call it.


ddebrunner:

PR #1031, opened 2017-05-11
Add ProcessingElement class to get a list of pe's from a Job

yvchan:
ProcessingElement class can also return metrics on itself, input & output ports as well.


ddebrunner:

PR #1053, opened 2017-05-23
[WIP] Initial Window class

ddebrunner:
Initial api for windowing in Python, currently limited to the invocation of SPL operators (e..g Aggregate).

Supports SPL sliding windows using the `last` method against a stream:

```
#  window of the last tuple on the stream
w = s.last()

# window of the last 20 tuples on the stream
w = s.last(20)

# window of the last five minutes
w = s.last(datetime.timedelta(minutes=5))
```

Example with trigger and SPL invocation:
```
agg = op.Map('spl.relational::Aggregate', s.last(4).trigger(2),
                    schema = 'tuple<uint64 sum, uint64 max>')
agg.sum = agg.output('Sum(seq)')
agg.max = agg.output('Max(seq)')
```

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -181,6 +181,7 @@ import os
 import time
 import inspect
 import logging
+import datetime
 from enum import Enum
 
 logger = logging.getLogger('streamsx.topology')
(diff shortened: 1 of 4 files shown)


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -720,6 +721,37 @@ def end_parallel(self):
         endP = Stream(self.topology, oport)
         return endP
 
+    def last(self, size=1):
```
wmarshall484: Maybe `last` could have an optional `when` parameter.

ddebrunner: What would 'when' do?

wmarshall484: It would specify when triggering should happen. Instead of:

```Python
s.last(4).trigger(2)
```

It could be

```
s.last(4, 2)
```

wmarshall484: On second thought, this doesn't keep a separation between the contents of the window and the details of its aggregation.

ddebrunner: yeah, though at the moment the Window class does have a default trigger of `count(1)` which also seems somewhat strange. It might be tied too much to SPL atm.

ddebrunner:
Merging with the expectation the api could change.

PR #1054, opened 2017-05-24
Cleanup the Instance object in the REST api

yvchan:
- remove instanceId from StreamsConnection 
 - add instanceId to StreamingAnalyticsConnection
  - remove InstancesArray.java as it's no longer needed
  - only expose get functions that are not likely to change in Instance.java (add in doc comments)


ddebrunner:

java/src/com/ibm/streamsx/rest/Instance.java
```diff
@@ -5,13 +5,20 @@
...
 import com.google.gson.annotations.Expose;
 
 /**
- * {@Instance}
+ * 
+ * An object describing an IBM Streams Instance
+ * 
+ */
+/**
+ * @author yvchan
```
ddebrunner: Please don't include '@author' tags.

PR #1057, opened 2017-05-26
add in better exception information for the REST API wrappers

yvchan:
Created a customized exception for REST errors so that we can identify what happened in the highlevels of code.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/RESTException.java
+++ b/java/src/com/ibm/streamsx/rest/RESTException.java
@@ -1,0 +1,113 @@
+/*
+# Licensed Materials - Property of IBM
+# Copyright IBM Corp. 2017
+ */
+
+package com.ibm.streamsx.rest;
+
+import java.io.IOException;
+
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonObject;
+import com.google.gson.annotations.Expose;
+
+/**
+ * Exception for REST api wrappers
+ */
+public class RESTException extends IOException {
+
+    private int status;
+    private RESTErrorMessage error;
+
+    /**
+     * Customized exception that can provide more information on REST errors
+     * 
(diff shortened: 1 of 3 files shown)


ddebrunner:

yvchan:
was just talking to John M about this code, and he's suggesting that i do not override getMessage() but set the message in the exception instead.  I'll make the change.

yvchan:
ok, the update is here...  added in a factory method to handle getting the streams error response back ensuring the getMessage() method returns a human readable string when it's set.

java/src/com/ibm/streamsx/rest/RESTException.java
```diff
@@ -0,0 +1,122 @@
...
+        return status;
+    }
+
+    /**
+     * Gets the IBM Streams message id for this exception
+     * 
+     * @return {@link String}
+     */
+
+    public String getStreamsErrorMessageId() {
```
ddebrunner: What's the purpose of this method?

java/src/com/ibm/streamsx/rest/RESTException.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        return id;
+    }
+
+    /**
+     * Gets the IBM Streams message for this exception
+     * 
+     * @return {@link String}
+     */
+    public String getStreamsErrorMessage() {
```
ddebrunner: What's the purpose of this method?

yvchan: i'll remove this, it isn't necessary.  a previous iteration did not set the message field the Exception, and this was to get the message from the error structure.  however, i re-worked this to always set the message, so this method is no longer required.

yvchan:
ok, the method has been removed and this should be ready to merge.

PR #1058, opened 2017-05-26
added javadoc items for the java rest api wrappers

yvchan:
in addition to the javadoc, i've also removed any functions that should be returning structures instead of strings.  (i.e, things like getHost() or getInstance() and have left them out until we need to add them in

### DIFF
--- a/java/src/com/ibm/streamsx/rest/Job.java
+++ b/java/src/com/ibm/streamsx/rest/Job.java
@@ -98,7 +98,7 @@ public class Job {
     /**
      * Gets a list of {@link Operator operators} for this job
      * 
-     * @return List of {@link Operator}
+     * @return List of {@link Operator IBM Streams Operators}
      * @throws IOException
      */
     public List<Operator> getOperators() throws IOException {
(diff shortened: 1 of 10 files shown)


ddebrunner:

java/src/com/ibm/streamsx/rest/Job.java
```diff
@@ -105,128 +109,137 @@ void setConnection( final StreamsConnection sc) {
...
+     * @return {@link List} of {@link String}
+     */
+    public List<String> getSubmitParameters() {
         return submitParameters;
     }
 
+    /**
+     * Gets the Epoch time when this job was submitted
+     * 
+     * @return long
```
ddebrunner: The `@return` needs to describe what is returned, not the type of the return.

ddebrunner:
General comment that the `@return` clause needs to describe what is returned, not the type. The type is already described by the method signature.

Look at existing javadoc (e.g. Java SDK) to see typical styles.

yvchan:
i've updated the @return clauses.

PR #1065, opened 2017-05-30
Add documentation for View object.

wmarshall484:


ddebrunner:
An example would be good, in the View overview.

wmarshall484:
Added: https://github.com/IBMStreams/streamsx.topology/pull/1065/files#diff-b43385a0128a584506883fa9e29996fdR917

PR #1076, opened 2017-06-09
remove unneccessary *Array classes from the code.

yvchan:
the Array classes were originally added in to facilitate the GSON decoding, but they are not neccessary, so this pull request is to remove them.


java/src/com/ibm/streamsx/rest/Operator.java
```diff
@@ -50,13 +51,27 @@
...
+        try {
+            opArray = new GsonBuilder().excludeFieldsWithoutExposeAnnotation().create().fromJson(operatorsList,
+                    OperatorArray.class);
+
+            opList = opArray.operators;
+            for (Operator op : opList) {
+                op.setConnection(sc);
+            }
+        } catch (IllegalStateException e) {
+            opList = new ArrayList<Operator>();
```
ddebrunner: Why is an empty list being returned on an IllegalStateException, why not just let the exception be thrown?

If an empty list is really needed, then `Collections.emptyList()` should typically be used.

yvchan: this is a runtime exception that can't really be acted upon by the user.  so by returning the empty list, most code will just iterate through, versus fail.  if it is a persistent problem, (i.e, the structure changed enough that Gson.fromJson() isn't working, there will be other failures that will trigger.   

i'll make the change to Collections.emptyList() as the return for all of these, instead of a new ArrayList.

ddebrunner: It just seems strange to hide errors, and to have extra code that is not really needed.

yvchan: hmm... my view has always been to fail gracefully where possible.  These are all the list retrieval operations that do this, not the individual ones.  In these cases, most users are likely to just iterate over the list, and move to the next level.  If the gson fails to parse, we won't trigger a full failure by throwing a runtime exception that users are not likely to have setup a catch for.   So for the common use case of just retrieving metrics of the various components and iterating through the list of objects, any object change will trigger a partial failure until a recompile is done.

yvchan:
will add one more commit to this, the exception i'm catching is the wrong one and should be com.google.gson.JsonSyntaxException

yvchan:
ok, last item is fixed.  this is ready to be merged.

PR #1093, opened 2017-06-15
Test java rest wrappers

yvchan:
Add in testcases for the java rest api's for both distributed and streaming analytics cases.  These have been placed into a test-unit called "unittest.restapi.base"

Includes a 2nd commit to remove the cancel jobs from the sample programs as they indiscriminately cancel the first job they find.

### DIFF
--- a/test/java/src/com/ibm/streamsx/rest/test/StreamingAnalyticsConnectionTest.java
+++ b/test/java/src/com/ibm/streamsx/rest/test/StreamingAnalyticsConnectionTest.java
@@ -1,0 +1,117 @@
+/*
+# Licensed Materials - Property of IBM
+# Copyright IBM Corp. 2017
+ */
+
+package com.ibm.streamsx.rest.test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeNotNull;
+import static org.junit.Assume.assumeTrue;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.junit.Test;
+
+import com.ibm.streamsx.rest.Instance;
+import com.ibm.streamsx.rest.RESTException;
+import com.ibm.streamsx.rest.StreamingAnalyticsConnection;
+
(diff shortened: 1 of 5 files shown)


ddebrunner:
Why don't the sample jobs cancel the job they submitted?

test/java/src/com/ibm/streamsx/rest/test/StreamingAnalyticsConnectionTest.java
```diff
@@ -0,0 +1,117 @@
...
+                String vcapServices = System.getenv("VCAP_SERVICES");
+
+                // if we don't have serviceName or vcapServices, skip the test
+                assumeNotNull(serviceName, vcapServices);
+
+                testType = "STREAMING_ANALYTICS_SERVICE";
+                connection = StreamingAnalyticsConnection.createInstance(vcapServices, serviceName);
+            }
+        } catch (IOException e) {
+            fail("This shouldn't cause an exception");
```
ddebrunner: don't hide exceptions - it makes resolving failures hard.

yvchan:
the sample programs don't actually submit a job, they just look at the system, and print the objects out if there are any jobs running.

test/java/src/com/ibm/streamsx/rest/test/StreamingAnalyticsConnectionTest.java
yvchan: the fail() is so that we explicitly fail the test if an exception is thrown.   we get the line number from the fail to show exactly where the test didn't behave properly.

there are a few cases where we expect an exception and we check the status code to ensure it's correct in that case.

ddebrunner: The test will fail due to an exception, but there will be not details on why it failed ...

Hiding exceptions is rarely a good idea in java - been burnt too many times on that

yvchan: ok.  i'll remove them.  give me a minute to add a new commit with that.

PR #1105, opened 2017-06-21
Add refresh to Element

ddebrunner:


java/src/com/ibm/streamsx/rest/Element.java
```diff
@@ -37,4 +42,31 @@ void setConnection(StreamsConnection connection) {
...
+    /**
+     * Refresh this element.
+     * 
+     * Attributes of this object are updated from current state
+     * though the Streams REST API.
+     * 
+     * @throws IOException Error communicating with Streams.
+     */
+    public void refresh() throws IOException {
+        String response = connection().getResponseString(self);
```
yvchan: not all objects fill in the "self" string, if we don't have one, we could end up throwing an exception (or we should just throw a specific "can't refresh this object" exception?

ddebrunner: Which objects don't fill in the self string?

yvchan: oddly enough, the Metrics object which is one that we'd like the refesh() to work on:
https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.restapi.doc/doc/restapis-metrics.html

based on what I'd seen, the only way to refresh the metrics objects will be to call the getMetrics() from the higher level object (i.e, Operator,  the Input/Output Ports, etc, ..)

ddebrunner: Yeah, I just added a commit that handles Metric. I should have said in the original PR that there was going to be a follow on commit.

ddebrunner: it actually uses the URI from the "parent" object, but appends the name of the metric as a query parameter:

`?name=my_metric_name`

That though is returned as a list of one item, so we have to manually unpack the values that we expect to change.

PR #1107, opened 2017-06-22
fix #1098: NPE if Java REST api response is 404.

### LINKED ISSUE #1098: NPE if the Java REST api response is 404.
I had a  bad URL and the response had a 404 with an empty string.

This lead to an NPE because error was null at:

https://github.com/IBMStreams/streamsx.topology/blob/master/java/src/com/ibm/streamsx/rest/RESTException.java#L49

Looks like Gson returned a null for error at line 37 because the JSON was at EOF.

yvchan:
the additions to RESTException.java prevents an NPE in the class.
the changes to StreamsConnection.java is to give a better error message on the exception specifically on a 404 error message that has no other information from IBM Streams...

### DIFF
--- a/java/src/com/ibm/streamsx/rest/StreamsConnection.java
+++ b/java/src/com/ibm/streamsx/rest/StreamsConnection.java
@@ -116,8 +116,14 @@ public class StreamsConnection {
             sReturn = EntityUtils.toString(hResponse.getEntity());
         } else if (HttpStatus.SC_NOT_FOUND == rcResponse) {
             // with a 404 message, we are likely to have a message from Streams
+            // but if not, provide a better message
             sReturn = EntityUtils.toString(hResponse.getEntity());
-            throw RESTException.create(rcResponse, sReturn);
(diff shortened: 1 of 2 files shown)


java/src/com/ibm/streamsx/rest/StreamsConnection.java
```diff
@@ -116,8 +116,14 @@ String getResponseString(String inputString) throws IOException {
             sReturn = EntityUtils.toString(hResponse.getEntity());
         } else if (HttpStatus.SC_NOT_FOUND == rcResponse) {
             // with a 404 message, we are likely to have a message from Streams
+            // but if not, provide a better message
             sReturn = EntityUtils.toString(hResponse.getEntity());
-            throw RESTException.create(rcResponse, sReturn);
+            if ((sReturn != null) && (!sReturn.equals(""))) {
```
ddebrunner: FYI - String has an `isEmpty()` method which is preferred to `equals("")`.

PR #1119, opened 2017-06-27
Fix 2.7 issues introduced with schema changes.

ddebrunner:


ddebrunner:
ran the 'test.python2' target successfully.

PR #1125, opened 2017-06-29
Simplify with list comprehensions

cclauss:
Also, use `or` to avoid other falsey values like "", [], {}, etc.


ddebrunner:
Thanks!

PR #1127, opened 2017-07-06
Improve naming for parallel regions.

### LINKED ISSUE #1002: Simplify parallel composite naming.

wmarshall484:
Users may now specify the composite name which prefixes operators in a parallel region. The default is now `parallel` instead of `paraComp`. Here we see a parallel region created with the name `CUSTOM`, and another with the default `parallel`.

```python
s = s.parallel(5, name="CUSTOM")
...
s = s.parallel(7)
...
```

newname

fixes #1002

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
@@ -109,7 +109,10 @@ class SPLGraph(object):
 
         if name is None:
             name = self._requested_name(None,action="Op", func = function)
-
+        
+        else:
+            name = self._requested_name(name, action="Op", func = function)
+            
         if(kind.startswith("$")):    
             op = Marker(len(self.operators), kind, name, {}, self)                           
         else:
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
@@ -109,10 +109,7 @@ class SPLGraph(object):
 
         if name is None:
             name = self._requested_name(None,action="Op", func = function)
-        
-        else:
-            name = self._requested_name(name, action="Op", func = function)
-            
+
         if(kind.startswith("$")):    
             op = Marker(len(self.operators), kind, name, {}, self)                           
         else:
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -684,6 +684,9 @@ class Stream(object):
         """
         if name is None:
             name = self.name
+            
+        name = self.topology.graph._requested_name(name, action='parallel', func=func)
+
         if routing == None or routing == Routing.ROUND_ROBIN:
             op2 = self.topology.graph.addOperator("$Parallel$", name=name)
             op2.addInputPort(outputPort=self.oport)
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -676,14 +676,14 @@ class Stream(object):
             func: Optional function called when :py:const:`Routing.HASH_PARTITIONED` routing is specified.
                 The function provides an integer value to be used as the hash that determines
                 the tuple channel routing.
-            name (String): The name to display for the parallel region.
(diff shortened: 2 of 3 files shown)


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -676,13 +676,16 @@ def parallel(self, width, routing=Routing.ROUND_ROBIN, func=None):
             func: Optional function called when :py:const:`Routing.HASH_PARTITIONED` routing is specified.
                 The function provides an integer value to be used as the hash that determines
                 the tuple channel routing.
+            name (String): The name to display for the parallel region.
```
ddebrunner: In Python it's `str`, not `String`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -676,13 +676,16 @@ def parallel(self, width, routing=Routing.ROUND_ROBIN, func=None):
...
 
         Returns:
             Stream: A stream for which subsequent transformations will be executed in parallel.
 
         """
+        if name is None:
+            name = "parallel"
```
ddebrunner: I wonder if the name should default to the name of the stream being parallelized?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
wmarshall484: Fixed.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
wmarshall484: Here's what it looks like when the stream name is used for the default parallel region name:

newname2

As you can see, the `map_lambda` operator carries the name into the parallel region.

wmarshall484: Personally, I think the default of `parallel` is more clear. Do you have a preference?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -109,7 +109,10 @@ def addOperator(self, kind, function=None, name=None, params=None, sl=None):
 
         if name is None:
             name = self._requested_name(None,action="Op", func = function)
-
+
```
ddebrunner: Why was this added, I'm not sure it's correct?

wmarshall484: Without this, SPL compilation fails when there are two operators of the same name.

wmarshall484:
```
SEVERE: t/parallel_test.spl:63:1: CDISP0025E ERROR: The following identifier is a duplicate: map_lambda_2.
Jun 30, 2017 11:30:07 AM com.ibm.streamsx.topology.internal.process.ProcessOutputToLogger run
SEVERE: t/parallel_test.spl:52:1: CDISP0032D DETAILS: The map_lambda_2 identifier was previously defined at this location.
Jun 30, 2017 11:30:07 AM com.ibm.streamsx.topology.internal.process.ProcessOutputToLogger run
SEVERE:
Jun 30, 2017 11:30:07 AM com.ibm.streamsx.topology.internal.process.ProcessOutputToLogger run
SEVERE: CDISP0092E ERROR: Because of previous compilation errors, the compile process cannot continue.
```

wmarshall484: If we have the following program:
```
topo = Topology("parallel_test")

s = topo.source(['1','2','3','4'])

s = s.parallel(5, name="CUSTOM").map(lambda x: -1*x).end_parallel()
s = s.map(lambda x: x)

s = s.parallel(5, name="CUSTOM").map(lambda x: -1*x).end_parallel()
s = s.map(lambda x: x)
```

It complains about the `CUSTOM` identifier being a duplicate.

ddebrunner: That's because the caller is meant to make the requested_name call, eg. see map, filter etc. in topology.py. Then the action correctly reflects what is being done, thus `_parallel` should make the call.

wmarshall484: Moved to `parallel` method.

PR #1129, opened 2017-07-06
Add layout note and initial use with hidden.

ddebrunner:
Adds an optional layout object to the JSON graph for an operator and indicates the hash adder/remover operators should be hidden for Java & Python.


wmarshall484:
Looks good. As an observation, the `layout` method in `_SPLInvocation` has different semantics than the `layout` method in `BOperatorInvocation`. The Python method acts like a setter, and the Java method acts like a getter.

ddebrunner:
yeah, mainly because of Java/Python differences, much easier to have a general purpose easy to use layout setter method in java than python, so in Java the caller deals with filling in the object.

Thanks for the review.

PR #1136, opened 2017-07-10
Windowing support.

wmarshall484:


wmarshall484:
Taking away the WIP tag. All tests pass, and the API is stable.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_h.cgt
```diff
@@ -46,6 +46,7 @@ if ($pyoutstyle eq 'dict') {
     // Members
     // Control for interaction with Python
     SplpyFuncOp *funcop_;
+    SplpyFuncOp *spl_in_object_out;
```
ddebrunner: I know this is a WIP but using two `SplpyFuncOp` instances is dangerous because it's more than a function pointer. The class represents an operator instance that is a functional operator, not just the function pointer. It would be better to either have an explicit PyObject* that is the second function pointer or extend `SplpyFuncOp` to represent operator with with two functions.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
```diff
@@ -103,22 +108,31 @@ print splpy_inputtuple2value($pystyle);
...
+
+      unsigned char const *data = value.getData();
+      unsigned char fmt = *data;
+      if (fmt == STREAMSX_TPP_PTR) {
+          __SPLTuplePyPtr *stp = (__SPLTuplePyPtr *)(data);
+          python_value = stp->pyptr;
+      }
+      // Anything ASCII is also Pickle (Python 2 default format)
+      else if (fmt <= STREAMSX_TPP_PICKLE) {
+          python_value = pySplValueToPyObject(value);
```
ddebrunner: Won't this result in the pickled bytes being inserted into the window, rather than the actual object?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
```diff
@@ -103,22 +108,31 @@ print splpy_inputtuple2value($pystyle);
...
+      }
+  <%} else {%>
+      python_value = streamsx::topology::Splpy::pyTupleMap(spl_in_object_out->callable(), value);
```
ddebrunner: I'm still unsure about using `pyTupleMap` here just because is seems to (almost) do the right thing.

Here' we are not *mapping* an object from one value to another, but just preserving the tuple in the window.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
wmarshall484: Fixed.

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
```diff
@@ -38,10 +38,20 @@ class SplpyFuncOp : public SplpyOp {
          addAppPythonPackages();
          loadAndWrapCallable(wrapfn);
       }
+
+    SplpyFuncOp(SPL::Operator * op, const std::string & wrapfn,
+		const std::string & module, const std::string & function) :
+         SplpyOp(op, "/opt/python/packages/streamsx/topology")
+      {
+         addAppPythonPackages();
```
ddebrunner: Isn't this duplicate code, e.g. the `addAppPythonPackages` and `loadAndWrapCallable(wrapfn)` will be called by the constructor for `SplpyOp` ?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -1285,6 +1285,47 @@ def trigger(self, when=1):
...
+        """
+        # WIP: TODO: support other window types and policies                                                                                                                                                                               
+
+        if self._config['evictPolicy'] != 'COUNT' or self._config['triggerPolicy'] != 'COUNT':
+            raise NotImplementedError("Currently, only windows with eviction policies of type COUNT and trigger policies of type COUNT are supported")
+
+        if schema is None:
+            schema = self.stream.oport.schema
+        
+        sl = _SourceLocation(_source_info(), "window")
```
ddebrunner: The action here is 'aggregate', not 'window'.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow.xml
```diff
@@ -0,0 +1,125 @@
+<?xml version="1.0" ?>
+<operatorModel
+  xmlns="http://www.ibm.com/xmlns/prod/streams/spl/operator" 
+  xmlns:cmn="http://www.ibm.com/xmlns/prod/streams/spl/common" 
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.ibm.com/xmlns/prod/streams/spl/operator operatorModel.xsd">
+  <cppOperatorModel>
+    <context> 
+      <description>Python Windowing Operation</description>
+      <iconUri size="16">../opt/icons/transform_16.gif</iconUri>
```
ddebrunner: This is an aggregate operator, so it should have aggregate icons.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
```diff
@@ -0,0 +1,253 @@
...
+using namespace streamsx::topology;
+
+<%SPL::CodeGen::implementationPrologue($model);%>
+
+@include "../pyspltuple.cgt"
+<%
+ # Select the Python wrapper function
+ my $pyoutstyle = splpy_tuplestyle($model->getOutputPortAt(0));
+ 
+ # Temporary, until more input types are supported. At the moment, only pass by reference types
```
ddebrunner: The code is somewhat confusing as comments indicate it only supports pass by reference, but there is code to support dict and pickled values. Maybe the code could be cleaned up to only include what is currently supported?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
```diff
@@ -0,0 +1,253 @@
...
+       out_wrapfn = "<%=$pybyrefwrapfunc%>";
+       occ_ = <%=$occ%>;
+    }
+<%
+    } 
+ }
+%>
+
+    funcop_ = new SplpyFuncOp(this, out_wrapfn);
+    spl_in_object_out = new SplpyFuncOp(this, in_wrapfn, functions_module, identity_function);
```
ddebrunner: This is possibly overkill. The wrapping capability was designed to call a user function, but in this case there is no need to call a user function (`identity`). The real requirement is to convert the raw Python object into the logical Python object of which there will be three cases:
         1. Python bytes to depicked object
         1. Python str to dict as JSON
         1. No need for conversion (Schema is String, Binary, SPL Schema (Python dict))`

With the wrapped user function case (e.g map), there is always a need to call a wrapping function which does optional conversion, calls the user function and handles the return. That's not true in this case.

Since this PR started, there is now an example of code having to call a Python function repeatably, in `MemoryViewCleanup` :
```
static PyObject * releaser = SplpyGeneral::loadFunctionGIL("streamsx.spl.runtime", "_splpy_release_memoryviews");
```

and calling it:

```
PyObject * ret = SplpyGeneral::pyCallObject(releaser(), args);
```



While it is implementation detail, it does end up adding a new `SplpyFuncOp` constructor that is not needed.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
```diff
@@ -0,0 +1,253 @@
...
+@include "../pyspltuple2dict.cgt"
+<%}%>
+
+  PyObject *python_value;
+
+// If the input style is pickle,
+{
+  // None of the streamsx::topology methods in this scope grab the lock
+  // so we need to do it here.
+  SplpyGIL lock;
```
ddebrunner: If the input is already a Python object are we grabbing the GIL for no reason?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow.xml
wmarshall484: Fixed.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
wmarshall484: Fixed. The GIL should now only be grabbed when necessary.

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_cpp.cgt
wmarshall484: pyTypleMap has been replaced with 'loadFunction()' and 'pyCallObject()

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/CCWindow/CCWindow_h.cgt
wmarshall484: Using a `SplpyFuncOp` has been restricted to wrapping user provided functions, of which there is only one for CCWindow. In the future, if there are two user provided functions for an operator, `SplpyFuncOp` could be extended to allow wrapping multiple user functions.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
wmarshall484: Fixed.

test/python/topology/test2_python_window.py
```diff
@@ -0,0 +1,85 @@
...
+        return time.localtime().tm_year - self.age;
+
+class TestPythonWindowing(unittest.TestCase):
+    def setUp(self):
+        Tester.setup_standalone(self)
+
+    def test_BasicCountCountWindow(self):
+        topo = Topology()
+        s = topo.source([1,2,3,4,5,6,7,8,9,10,11,12,13,14,15])
+        s = s.last(10).trigger(1).aggregate(lambda x: sum(x)/len(x))
```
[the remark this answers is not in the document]
wmarshall484: Changed several trigger values.

wmarshall484:
Notes/Changes:

* The `dict` type doesn't work on either input or output. Compilation reports an error and exits if this scenario is encountered.

* I've updated the code to use `loadFunction` and `pyCallObject` to invoke non user-provided functions such as the `streamsx.runtime.identity` function.

* The CCWindow's aggregate function is still wrapped in a SplpyFuncOp since it is provided by a user.

* In the event that the input spl type does not wrap a reference or a pickled python object, `pySplProcessTyple` is used to both convert the spl tuple type, and call the non user-provided function.

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
```diff
@@ -83,6 +85,23 @@ class SplpyFuncOp : public SplpyOp {
                "streamsx.topology.runtime", wrapfn, appCallable, NULL));
       }
 
+
```
wmarshall484: TODO remove

ddebrunner:
+1 to merge as-is assuming an issue is created to track follow on implementation tasks for this PR.

PR #1210, opened 2017-09-07
Support SPL decimal32/64/128 in Python as decimal.Decimal

ddebrunner:
Fixes #1210


ddebrunner:
Cherry-picked to v1_7: 6316a3effa6a15fbaf64a604e2ac84418bc0594f

PR #1216, opened 2017-09-08
Remove addSPLOperator with broken name generator.

### LINKED ISSUE #1215: JCPTest failing on v1_7 and master.

ddebrunner:
Fixes #1215 - name was being incorrectly calculated (`+2`) was inside `lastIndexOf()` rather than adding two to the returned index. Only used in one place, so removed and have `addJobControlPlane` explicitly set the name.


ddebrunner:
Cherry-picked to v1_7: 6316a3effa6a15fbaf64a604e2ac84418bc0594f

PR #1218, opened 2017-09-10
Change approach for mapping names to SPL identifiers

### LINKED ISSUE #1212: Operator names that exceed 255 characters cause Streams compilation to fail (due to Linux file name being too long)
I have an application that generated operator names and when topology invoked the Streams compiler, I got the following error:
```
 CDISP0105E ERROR: The following file cannot be created: ...   The error is: file name is too long.
```

### LINKED ISSUE #1217: Build identifiers with non-ascii characters can cause 404/500 errors with build service.

ddebrunner:
Maps each name that is not a SPL identifier OR over 80 characters to a md5 hash based name from the bytes of the original name.

Fixes #1212 

Uses the SPL mapped identifier of the app name as the base for the build identifier to avoid any problematic names.
Fixes #1217

### DIFF
--- a/java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
@@ -21,6 +21,9 @@ import static com.ibm.streamsx.topology.internal.gson.GsonUtilities.jstring;
 
 import java.io.IOException;
 import java.math.BigInteger;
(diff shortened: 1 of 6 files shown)


ddebrunner:
Cherry-picked to v1_7: 
- 1a7e66fd035f3db199026dbbea66bb7ceb9aba30
- 943f4e9770c409ffe332569a14c26a48674a6f02

PR #1222, opened 2017-09-13
Delay log/trace setup until PYTHONPATH is set

ddebrunner:


ddebrunner:
Cherry-picked to v1_7: dc18deb79bb30befc08cc729353f5224116a6efd

PR #1223, opened 2017-09-13
Fix test when VCAP_SERVICES is the raw JSON

ddebrunner:


ddebrunner:
Cherry-picked to v1_7; 5ede34dc098189af7fb6454b091add19cc4d488a

Issue #1231, opened 2017-09-21
Stream.map from a Json to Json schema fails. 


ddebrunner:
`s.as_json().map(..., schema=CommonSchema.Json)`

fails as there is no `json_in_json_out` function.

ddebrunner:
Fixed.

PR #1232, opened 2017-09-21
Add object mapping functions for Python functional operators.

### LINKED ISSUE #1230: Stream.map from a String to String schema fails.
`s.as_string().map(..., schema=CommonSchema.String`

fails as there is no `string_in_string_out` function.

ddebrunner:
- Adds object mapping like `object_in_object_out`
- Changes the _style_ functions to be variables that reference the required callable class either directly or through another variable
- Moves the _style_ variables to a single location which exposed #1230 and #1231
- Add additional tests to cover all the combinations where a schema can be changed.
- Fixes #1230 #1231


ddebrunner:
Cherry-picked to v1_7
6cf890f6bc2c2a51963062d9702423ef1b7cbb6c
a92458c0450c49add556f0ec2bb0f17332860fa5

and v1_7 bumped to 1.7.4.

Issue #1235, opened 2017-09-21
Support passing structured schemas as Python tuples.


ddebrunner:
Currently a structured schema is passed into a Python callable as a dict:

`tuple<rstring a, int32 b>` is passed as `{'a': 'hello', 'b': 42}`

Allow passing as a

- [x] `tuple` - would be passed as `('hello', 42)`
- [x] `collections.namedtuple` - would be passed as a named tuple, e.g. `t.a == 'hello'` and `t.b == 42`
- [ ] `typing.NamedTuple` - would be passed as a named tuple, e.g. `t.a == 'hello'` and `t.b == 42` and include type hints (Python 3.5 only).
- [ ] specific named tuple - A named tuple class provided by the user.

Concept is to have additional methods on `StreamSchema` to indicate how the tuples should be handled, e.g.
```
   s.map(myfunc, schema=StreamSchema('tuple<rstring a, int32 b>').as_tuple())
```

ddebrunner:
@chanskw FYI plus the example in PR #1236

ddebrunner:
named tuple support added in 1.9 under #1376

markheger:
supported

PR #1236, opened 2017-09-22
Passing as tuple support

ddebrunner:
**updated** Now supports `map`, `filter`, `for_each`, `parallel (hash)`, `flat_map`.

Allows a `StreamSchema` to be modified to indicate for a structured stream that its stream tuples are to be passed as instances of `tuple` instead of `dict`.

```
# Existing/default - values passed as dict objects with the attribute names being the keys. 
sd = StreamSchema('tuple<rstring id, float64 value>')

# s1 is a a stream with the schema 'tuple<rstring id, float64 value>' and
# any processing against the stream will have stream tuples passed as dict objects
# e.g. {'id': 'Sensor', 'value': 23.2}
s1 = s.map(fm, schema=sd)

# So fs will be called as fs( {'id': 'Sensor', 'value': 23.2} )
def fs(t):
    print("Sensor:" + t['id'] + '=' + str(t['value']))
s1.for_each(fs)
```

Now by calling `as_tuple()` against the `StreamSchema` object a stream can be created where its stream tuples will be processed as tuple objects.

```
# New - values passed as tuple objects with the values being in order of the schema
sd = StreamSchema('tuple<rstring id, float64 value>').as_tuple()

# s1 is a a stream with the schema 'tuple<rstring id, float64 value>' and
# any processing against the stream will have stream tuples passed as tuple objects
# e.g. ('Sensor', 23.2)
s1 = s.map(fm, schema=sd)

# So callable fs will be called as fs( ('Sensor', 'value': 23.2) )
def fs(t):
    print("Sensor:" + t[0] + '=' + str(t[1]))
s1.for_each(fs)
```

@wmarshall484 Review requested for the api changes mainly, `as_tuple`, `as_dict` and `style` methods on `StreamSchema`.

WIP for #1235 - only regular `tuple` support.

### DIFF
--- a/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.cgt
+++ b/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.cgt
@@ -19,6 +19,11 @@
 
  # determine which input tuple style is being used
 
- my $pystyle = splpy_tuplestyle($model->getInputPortAt(0));
+ my $pystyle = $model->getParameterByName("pyStyle");
+ if ($pystyle) {
+     $pystyle = substr $pystyle->getValueAt(0)->getSPLExpression(), 1, -1;
+ } else {
+     $pystyle = splpy_tuplestyle($model->getInputPortAt(0));
+ }
 %>
 
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
@@ -37,7 +37,7 @@ def is_common(schema):
         return True
     if isinstance(schema, str):
         return is_common(StreamSchema(schema))
-    raise TypeError()
+    return False
 
 # Parses a schema of the form 'tuple<...>'
 # _parse returns a list of the schema attributes,
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
@@ -21,6 +21,24 @@ import token
 import tokenize
 
 
+def is_common(schema):
+    """
+    Is `schema` an common schema
+    Args:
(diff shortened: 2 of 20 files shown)


com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.cgt
```diff
@@ -19,6 +19,11 @@
 
  # determine which input tuple style is being used
 
- my $pystyle = splpy_tuplestyle($model->getInputPortAt(0));
+ my $pystyle = $model->getParameterByName("pyStyle");
+ if ($pystyle) {
+     $pystyle = substr $pystyle->getValueAt(0)->getSPLExpression(), 1, -1;
```
wmarshall484: Does the substring here remove quotes?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
```diff
@@ -21,6 +21,24 @@
...
+        bool: ``True`` if schema is a common schema, otherwise ``False``.
+
+    """
+    if isinstance(schema, StreamSchema):
+        return schema.schema() in _SCHEMA_COMMON
+    if isinstance(schema, CommonSchema):
+        return True
+    if isinstance(schema, str):
+        return is_common(StreamSchema(schema))
+    raise TypeError()
```
wmarshall484: Rather than raise a TypeError, shouldn't this just return `False`?

com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/pyspltuple.cgt
ddebrunner: yes

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/schema.py
ddebrunner: I'll change that. Thanks,

PR #1252, opened 2017-10-02
Fix pending schemas with pystyle

ddebrunner:


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -284,11 +284,17 @@ def generateSPLOperator(self):
...
+
         # Add parameters as their string representation
         # unless they value has a spl_json() function,
         # then use that
         _params = {}
+
+        # Fix up any pending streams for input style
+        if 'pyStyle' in self.params and 'pending' == self.params['pyStyle']\
+                and self.kind.startswith('com.ibm.streamsx.topology.functional.python'):
+            StreamSchema._fnop_style(self.inputPorts[0].schema, self, 'pyStyle')
```
wmarshall484: Does this mean the pending streams turn into nops?

ddebrunner: No - fnop means _function (fn) operator (op)_.

This is fixing up the tuple style for downstream operators of a 'pending' stream. When the operator is invoked (through something like a `s.map()` or an SPL operator invocation) we can't determine the style from the schema as it is not known (since the stream is pending). Thus we set the style to `pending` and then fix up at json generation time when we will know the schema.

PR #1259, opened 2017-10-04
Aggregate operator now uses code gen API to obtain window definition.

wmarshall484:


wmarshall484:
No new functionality is added. Time based windows and time based tests will be added in a fourthcoming PR.

PR #1260, opened 2017-10-04
Fix documentation for aggregate operator.

### LINKED ISSUE #1258: Improve Window.aggregate documentation
- [ ] Summary sentence is incorrect, it does not "declare a function ..."
- [ ] Need to state order of tuples in the list
- [ ] `name`, `schema` parameters not defined
- [ ] Code example is not in a code block (compare to code example in the Window class documentation.

wmarshall484:
Fixes #1258

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
@@ -1309,21 +1309,25 @@ class Window(object):
         return tw
 
     def aggregate(self, function, name=None, schema = None):
-        """Declares a function or callable to aggregate the contents of 
-        the window when it is triggered.
+        """Aggregates the contents of the window when the window is
+        triggered.
         
-        The supplied function is passed a list containing the contents 
-        of the window. The return values of the function are passed as
-        the tuples on the returned `stream`. For example, a window that
-        calculates a moving average of the last  10 tuples could be 
-        written as follows:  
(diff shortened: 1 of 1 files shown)


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -1309,21 +1309,25 @@ def trigger(self, when=1):
...
         
-        The supplied function is passed a list containing the contents 
-        of the window. The return values of the function are passed as
-        the tuples on the returned `stream`. For example, a window that
-        calculates a moving average of the last  10 tuples could be 
-        written as follows:  
+        Upon a window trigger, the supplied function is passed a list containing 
+        the contents of the window: ``function(items)``. The order of the window 
+        items in the list are the order in which they were each received by the 
+        window. The return values of the function are passed as the tuples on
```
ddebrunner: _The return values of the function are passed as the tuples on the returned `stream`_

Can be read as though multiple tuples can be submitted for a single aggregation.

Look at how `Stream.map` defines how a function is called and use  the same style.

wmarshall484: Update to:
```
If the function's return value is not `None` then the result will
be submitted as a tuple on the returned stream. If the return value is 
`None` then no tuple submission will occur.
```

wmarshall484: See #1261

PR #1285, opened 2017-10-26
Support retrieving the application logs via the REST API.

wmarshall484:


com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -395,6 +400,47 @@ class Job(_ResourceElement):
...
+        if name is None:
+            # Take the job name and remove colons (colons confuse the unix 'tar' command)
+            name = ''.join(self.name.split(':'))
+
+        name = name + "_app_logs.tar"
+        
+        if path is None:
+            path = os.getcwd()
+
+        path = path + "/" + name
```
ddebrunner: Use `os.path.join`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -395,6 +400,47 @@ class Job(_ResourceElement):
...
+
+         Returns:
+            str: the path to the application logs tar file.
+         """
+        logger.debug("Retrieving application logs from: " + self.applicationLogTrace)
+        logs = self.rest_client.make_raw_request(self.applicationLogTrace)
+        
+        if name is None:
+            # Take the job name and remove colons (colons confuse the unix 'tar' command)
+            name = ''.join(self.name.split(':'))
```
ddebrunner: Can you include the job id, so that multiple runs of a job with the same name produce different files?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -395,6 +400,47 @@ class Job(_ResourceElement):
...
+
+        name = name + "_app_logs.tar"
+        
+        if path is None:
+            path = os.getcwd()
+
+        path = path + "/" + name
+        try:
+            logfile = open(path, 'w+b')
+            logfile.write(logs.content)
```
ddebrunner: Is it possible to stream from the HTTP response into the file or does that do this?

Logs can be big and having to load it into memory first could be slow and problematic.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -395,6 +400,47 @@ class Job(_ResourceElement):
         >>> print (jobs[0].health)
         healthy
     """
+    def get_application_trace(self, path=None, name=None):
```
ddebrunner: It's a little inconsistent that the method is `get_application_trace`, the docstring says `application log` and the file name suffix is `app_logs.tar`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -395,6 +400,47 @@ class Job(_ResourceElement):
...
+        """Retrieves the application log and saves it to the specified path with the given name as a tar file.
+
+        If logs are retrieved with the same path and name as previously retrieved logs, the prior logs will be
+        overwritten.
+
+        Attributes:
+            path (str): a valid directory in which to save the application log output. Defaults to current dir.
+            name (str): the filename of the created tar file. Defaults to a name based on the job name.
```
ddebrunner: The code doesn't create a file with `name` when it is specified, it appends `_app_logs.tar`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: We should be consistent within this definition and consistent with existing Streams doc, e.g.:

https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.jmxapi.doc/api/com/ibm/streams/management/job/JobMXBean.html#retrieveApplicationLogAndTraceFiles-java.lang.String-

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Fixed.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Added to the name.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: It now streams the tar file contents using a 1K chunk size.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Fixed. Should all be "application logs" now.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Updated to `prefix` from `name`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: I'm fine with the naming, but the doc strings is not consistent with the JMX description, specifically: 
_ all application log and trace files for the job in tar archive format._

ie. no mention of trace or what format the resultant file is.

wmarshall484: > Returns an HTTPS URI to retrieve all application log and trace files for the job in tar archive format.

Okay, to be consistent with the JMX doc I added two pieces of information:
* The trace information is also retrieved 
* The retrieved logs are from the associated job
The doc had already mentioned it's saved as a tar file.

wmarshall484:
SAS rest test now verify that the application logs are retrieved.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -395,6 +404,49 @@ class Job(_ResourceElement):
...
+            
+        prefix = prefix + "_" + self.id
+        name = prefix + "_app_logs.tar"
+        
+        if path is None:
+            path = os.getcwd()
+
+        path = os.path.join(path, name)
+        try:
+            logfile = open(path, 'w+b')
```
ddebrunner: Should really use a `with` clause to ensure the file is close on error

ddebrunner:
cherry-picked to v1_8

PR #1292, opened 2017-11-08
Fix get_application_logs pythondoc

ddebrunner:
- Fix the parameters to be documented as arguments and not attributes
- Fix indentation of the Returns clause (caused it not to show up)
- add a `versionadded` tag.


ddebrunner:
cherry-picked to v1_8

PR #1294, opened 2017-11-08
Fix #1293 - Support functions for@spl.source

### LINKED ISSUE #1293: Functions not supported for @spl.source
A decorated function like this produces a `TypeError`.
```
@spl.source()
def range200():
    return zip(range(200))
```

ddebrunner:


ddebrunner:
Cherry-picked to v1_8

PR #1296, opened 2017-11-10
Add submit_job() method to Python wrapper for Streams REST API

ejpring:


com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1182,45 @@ def __init__(self, rest_client, credentials):
     def _get_url(self, req_name):
         return self._credentials['rest_url'] + self._credentials[req_name]
 
+    def submit_job(self, sab_file, application_name,  parameters=[], configuration={}, tracing='info' ):
```
ddebrunner: Having mutable lists/dicts as the defaults is the incorrect approach for python, as the same object is used for all calls. Instead the default should be `None`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1182,45 @@ def __init__(self, rest_client, credentials):
...
+        """Submit a compiled Streams Application Bundle (a SAB file) to run in this Streamin Analytics service.
+        
+        Args:
+        sab_file (string): path to the compiled SAB file containing the application to be submitted
```
ddebrunner: The format of the args needs to be in Google style using Python types, e.g.

`sab_file(str)`

`configuration(dict)`  - no need to say it's optional. that's implied from the signature.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1182,45 @@ def __init__(self, rest_client, credentials):
...
+        application_name (string): name of the application in the SAB file
+        parameters (list of {name/value} pairs, optional): application's submission-time parameters
+        configuration (dict, optional): hash of deployment configuration parameters 
+        tracing (string, optional): one of 'off', 'error', 'warn', 'info',' debug', or 'trace', defaults to 'info'
```
ddebrunner: Elsewhere we use Python tracing levels to indicate the level, can it match the existing use.

Also the default should be `None` to use the instance default.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1182,45 @@ def __init__(self, rest_client, credentials):
     def _get_url(self, req_name):
         return self._credentials['rest_url'] + self._credentials[req_name]
 
+    def submit_job(self, sab_file, application_name,  parameters=[], configuration={}, tracing='info' ):
+        """Submit a compiled Streams Application Bundle (a SAB file) to run in this Streamin Analytics service.
+        
+        Args:
+        sab_file (string): path to the compiled SAB file containing the application to be submitted
+        application_name (string): name of the application in the SAB file
```
ddebrunner: Not sure why we need to know the application name.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1182,45 @@ def __init__(self, rest_client, credentials):
...
+        parameters (list of {name/value} pairs, optional): application's submission-time parameters
+        configuration (dict, optional): hash of deployment configuration parameters
```
ddebrunner: I think we need to design the api to support job config overlays fully rather than just a subset.
We probably have at least these cases:

- Pass in a complete JCO file/JSON - allows full flexibility.
- Pass in a JobConfig object which is the Python binding to a JCO.

Setting individual items might be better handled through the JobConfig object instead of a dict of key value pairs.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1182,45 @@ def __init__(self, rest_client, credentials):
...
+        parameters (list of {name/value} pairs, optional): application's submission-time parameters
+        configuration (dict, optional): hash of deployment configuration parameters 
+        tracing (string, optional): one of 'off', 'error', 'warn', 'info',' debug', or 'trace', defaults to 'info'
+        
+        For details of job configuration overlays, see:
+        https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.admin.doc/doc/job_configuration_overlays.html
+        https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.ref.doc/doc/submitjobparameters.html
+        
+        Returns:
+        dict: JSON response with name of submitted job
```
ddebrunner: To match the style of the REST api it should return a `Job` instance, not the raw JSON.

ddebrunner:
We also really require tests with new features, the existing rest rests are under `test/python/rest`

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Agreed -- I've changed the method accordingly.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Agreed -- I've changed the method accordingly.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: We don't, so I've removed it.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Agreed -- I've moved the 'target_pe_count' and 'submission_parameters' arguments to the JobConfig constructor, and changed the submit_job() method to optionally accept a job configuration overlay.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Agreed -- I've updated the submit_job() method to pass 'tracing' via an optional job config overlay.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Not sure this is a good idea. All other StreamingAnalyticsService methods return the JSON response from the service.

test/python/rest/README.md
```diff
@@ -2,17 +2,21 @@
...
-        unittest.assertEquals("asdf".upper(), "ASDF")
+        logger.debug("Beginning test: test_upper")
+        self.assertEqual("asdf".upper(), "ASDF")
 ```
 
 ## Running tests
 To run the test suite, just invoke
 ```
-python -m unittest discover
+python -m unittest -v string_tests
```
ddebrunner: Why was this changed?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -546,6 +574,22 @@ def cancel(self, force=False):
             return True
         raise NotImplementedError('Job.cancel()')
 
+    def store_logs(self, filename=None):
```
ddebrunner: How does this relate to `get_application_logs` ?

ddebrunner:
Can we keep a pull request focused on adding a single piece of functionality, otherwise it can take forever for a pr to be merged as new change that adds info needs additional review, and then if the subsequent changes adds new stuff yet another review is needed.

branches are cheap, recommended to keep unrelated changes in separate branches and prs.
(e.g. making changes in your master is not recommeded).

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: The store_logs() method uses the same REST API as get_application_logs(). The difference in the signature is that it has only one optional argument, which allows the caller to specify the tarball's full path. The difference in the implementation is that all of the download code is shared with get_console_log() and get_application_trace(). We certainly don't need both store_logs(), but I really don't think the additional characters get_application_logs() inserts into the filename are a good idea.

test/python/rest/README.md
ejpring: Oops. I had intended to show how to run all tests vs. just one test, but never finished that update. Sorry about that.

ejpring:
I could put each change in a separate branch, I suppose, but I think of all these changes as a single functional unit, and the unittest depends on all of the new methods. I wrote all of this code last month while I was working with the WDS folks in Pittsburgh. Their need is for APIs that allow a programmer to step through a complete life cycle during development, and then automate the whole cycle for testing and deployment. I'm still missing the 'remote build' step, so I'll put that in a separate branch ...

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: We need a single mechanism in the api, with consistent naming, we don't just want to add alternatives to support a different style.

ddebrunner: Can you separate out this multiple changes into isolated prs. Submit job is separate functionality to any change in the `get_application_logs` api.

ejpring: Sure. I've replaced this pull request with two more, one for submit_job() and one for store_logs()

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -405,15 +435,13 @@ class Job(_ResourceElement):
         healthy
     """
     def get_application_logs(self, path=None, prefix=None):
-        """Retrieves the application log and trace files of the job and saves them as a tar file to the specified path
-        with the given name.
+        """Retrieves the console logs and application traces of the job's processing elements (PEs) and stores them in a compressed 'tar' archive (that is, a '.tar.gz' file) with the specified file name in the specified directory path.
```
wmarshall484: Nit: try to keep documentations lines less than 80 characters. It affects how documentation is displayed in editors like PyCharm.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -546,6 +574,22 @@ def cancel(self, force=False):
             return True
         raise NotImplementedError('Job.cancel()')
 
+    def store_logs(self, filename=None):
+        """Download the console logs and application traces from a job and store them locally in a compressed 'tar' archive (that is, a '.tar.gz' file).
+        
+        Args:
+        filename (str): directory/file path of a compreseed 'tar' archive where the logs and traces should be stored. The default value is 'name.tar.gz' in the current directory, where 'name' is the job name.
```
wmarshall484: "compressed"

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -546,6 +574,22 @@ def cancel(self, force=False):
...
+        
+        Returns:
+        str: directory/filename of the compressed 'tar' archive where the logs and traces were stored
+        
+        Exceptions:
+        HTTP GET failed, error xxx xxxxxxxxxxxxx from URL xxxxxxxxxxxxxxxx
+        HTTP GET expected response of type application/x-compressed, got xxxxx/xxxxxx, from URL xxxxxxxxxxxxxx
+        """
+        
+        return self._store_file( ( filename if filename else self.name+'.tar.gz' ), self.applicationLogTrace, 'application/x-compressed')
```
wmarshall484: The job name contains colons, for example `strings::PyStringWindowTest.tar.gz` would be the output of the logs of the string window test.

The tar command gets confused by the presence of colons while extracting:
```
[streamsadmin@streamsqse]$ tar -xvzf strings::PyStringWindowTest_1000_10_0_app_logs.tar
tar (child): Cannot connect to strings: resolve failed

gzip: stdin: unexpected end of file
tar: Child returned status 128
tar: Error is not recoverable: exiting now
```

In `get_application_logs`, the colons are simply removed to prevent this. I could be convinced that we should keep the colons in the file name, but we should be consistent with our file names at the very least.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1264,32 @@ def __init__(self, rest_client, credentials):
     def _get_url(self, req_name):
         return self._credentials['rest_url'] + self._credentials[req_name]
 
+    def submit_job(self, sab_file, configuration=None):
+        """Submit a compiled Streams Application Bundle (a SAB file) to run in this Streamin Analytics service, optionally with a job configuration overlay.
```
wmarshall484: "Streaming"

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: I'm not familiar with PyCharm. I've reformatted the method documentation to 80-character lines. If that messes up the paragraph flow in generated HTML, let me know and I'll try something different.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Got it -- thanks.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Good point. I've noted the problem the 'tar' command has with filenames containing colons in the method documentation, and suggested a couple of workarounds.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: Got it -- thanks.

PR #1297, opened 2017-11-14
[WIP] Streaming Analytics Service v2 support, basic functions but not final.

johnmac-ibm:
Old direct uses of the Streaming Analytics Service REST API (mostly via
`RestUtils.java`, and use of responses) and the APIs added for Beam runner
in `com.ibm.streamsx.rest` have been changed to use factory methods to
create version-specific implementations.

Tested against the current SASv2 implementation in stage1, and current v1
services using Beam submit of `.sab` and remote build and submit, also with
collection of metrics via Streams REST API.

Not final, still undergoing testing.

For now, to test against a v2 instance, you will need to manually add `v2_rest_url`
and `apikey` to credentials, and have to generate an API key in Bluemix. The
`v2_rest_url` URL is of the form:

`https://streams-app-service-test.stage1.ng.bluemix.net/v2/streaming_analytics/<id>`

where `<id>` is the instance ID, which you can get from the Dashboard URL.

The above is required as a minimum, but some further improvements are
possible:

`RestUtils` and the `streamsx.rest` package had some similar code and error
handling that should be cleaned up and made common.

`RemoteStreamingAnalyticsServiceStreamsContext` is still aware of job ID
member name in responses, this should be pushed down into the version-specific
implementations, but this may require some changes to some of the contexts.

The construction of `StreamsConnection` implementations are not ideal because
2 of 3 start with URLs near the root, and common code finds the instances
from that, but v2 starts with the instance, constructs the root and forgets
the instance, which is then reconstructed by the common code. In particular
it would have problems if / when a subsequent version starts supporting more
than one instance per Streaming Analytics Service instance.

Authentication and SAS version are currently coupled, in practice and the
code, but conceptually authentication could be separated out into.

If SAS stabilizes on having a version-specific path prefix (current /v2,
next /v3, etc.) and the basic starting elements remain similar, the factory
might be able to load implementations by constructing the class name and
loading it, instead of switching on known versions.


java/src/com/ibm/streamsx/rest/StreamsConnection.java
```diff
@@ -190,64 +33,13 @@ String getResponseString(String inputString) throws IOException {
...
-        }
-        return allowInsecureHosts;
-    }
+    Instance getInstance(String instanceId) throws IOException;
 
     /**
      * Cancels a job at this streams connection identified by the jobId
+     * FIXME: Does this really belong at this level or should it be in Instance
+     * or even Job? I think the current implementation only works on the default
+     * domain/instance
```
pgerv12: +1 It doesn't make sense to have it here.

java/src/com/ibm/streamsx/topology/internal/context/service/RemoteStreamingAnalyticsServiceStreamsContext.java
```diff
@@ -39,10 +39,13 @@ public RemoteStreamingAnalyticsServiceStreamsContext() {
...
-        
+
         JsonObject results = object(entity.submission, RemoteContext.SUBMISSION_RESULTS);
         if (results != null) {
-            String jobId = jstring(results, "jobId");
+            // V2 uses id, V1 uses jobId
+            String jobId = jstring(results, "id");
+            if (jobId == null)
+                jobId = jstring(results, "jobId");
             if (jobId != null)
```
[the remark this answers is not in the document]
johnmac-ibm: Ideally, I think this whole chunk should move down into the version-specific implementation, but that looked like it would require more changes to the context than I really wanted at this time.

As it is, it's looking for one of two members but doesn't know which. It could be restructured of course, or even check with `has` before fetching, but it wouldn't be correct to just drop an `else` in before the second `if` because we want that code to execute if `jobId` got set in the first `if`.

ddebrunner:
Won't this break binary compatibility by changing items like `StreamsConnection` to an interface from a class?

johnmac-ibm:
@ddebrunner It definitely will. As far as I know the Runner for Apache Beam is the only current user of this API.

I did ask about this at the outset in a scrum and was under the impression it was okay, but if I misunderstood I can revert to classes that just always forward to a delegate.

I'm not certain I can maintain all the old public methods though (I'm thinking mostly of `allowInsecureHosts`, since it affects the state of the connection after the delegate would have been constructed), but I can take a stab at it.

ddebrunner:
We published this api as part of the 1.7 release, thus we don't know who might be using it.

johnmac-ibm:
@ddebrunner I have a patch that I think maintains binary and source compatibility (though deprecates those classes), but the test v2 environment is unavailable until tomorrow morning, so I can't fully test it yet.

Does topology have any specific mechanisms for backwards compatibility testing, eg. running samples or tests compiled as part of an older version against the new build?

wmarshall484:
@johnmac-ibm, have you tried running the `StreamsConnectionTest` and `StreamingAnalyticsConnectionTest`? They depend on having a `StreamsConnection` class.

https://github.com/IBMStreams/streamsx.topology/tree/master/test/java/src/com/ibm/streamsx/rest/test

java/src/com/ibm/streamsx/rest/StreamsConnection.java
wmarshall484: I agree, I think this should be moved to `Job`.

johnmac-ibm: For binary and source compatibility I'll have to leave it in the `StreamsConnection` class, but I'll remove it from the new interface, and add it to `Job`.

johnmac-ibm:
I have updated to restore backward compatibility (though I deprecated the old classes), and tested against the old versions of `StreamsConnectionTest` and `StreamingAnalyticsConnectionTest` (with slight modifications described below) as suggested by @wmarshall484 

- additional info in some asserts
- change due to behaviour change of `cancelJob`: it can return `true` / `false`, but also declares it can throw, and did not specify what it did for an invalid job. It used to return `false` and the test depended on that, it now throws, which I like better so I tweaked the test slightly. I can force it to return `false` in this case if reviewers prefer.

In my QSE, test results with this pull request match master, but 2 tests fail the same way in both, apparently due to changes in names of operators by topology on master. Someone more familiar with topology should feel free to propose the best way to fix the tests. :-)

One final caveat is that the old APIs cannot currently test against the V2 REST APIs without tweaking the source, because the old APIs have no mechanism to create a `StreamingAnalyticsConnection` with a service that has invalid certs like the current test environment does.

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnectionInterface.java
```diff
@@ -0,0 +1,30 @@
...
+# Copyright IBM Corp. 2017
+ */
+package com.ibm.streamsx.rest;
+
+import java.io.IOException;
+
+/**
+ * The initial interface for interacting with the REST API of a Streams instance
+ * on the Streaming Analytics Service on IBM Bluemix. Since Streaming Analytics
+ * Service has a single Streams instance (at least for now), this extends
```
ddebrunner: Don't put in comments that imply possible future changes.

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnectionInterface.java
```diff
@@ -0,0 +1,30 @@
...
+ * {@link StreamsConnectionInterface} and adds a convenience method {@link getInstance}
+ * to get the single instance.
+ * <p>
+ * Concrete instances are created with the factory methods in
+ * {@link StreamsRestFactory}
+ */
+public interface StreamingAnalyticsConnectionInterface extends StreamsConnectionInterface {
```
ddebrunner: Should we use the eclipse style here? `IStreamingAnalyticsConnection` instead of `StreamingAnalyticsConnectionInterface`.

java/src/com/ibm/streamsx/rest/StreamsConnection.java
```diff
@@ -6,114 +6,35 @@
...
-                    remants = iUrl.substring(index + keyword.length(), iUrl.length());
-                }
-                // verify that there's nothing at the end
-                if (remants.equals(endString)) {
-                    // make sure we stop at streams/rest
-                    this.url = iUrl.substring(0, index + keyword.length());
-                }
-            }
-        }
+    protected StreamsConnection(StreamsConnectionInterface delegate,
```
ddebrunner: Since the existing `StreamsConnection` constructor was protected, it is part of the public api so it needs to remain.

Maybe make this a new constructor and have it package protected.

johnmac-ibm: The constructor won't be a problem, but there are some other protected member fields that while I could leave in, almost certainly won't work as they used to, eg. the `apiKey` won't be usable in the same way it used to be (it's changed in the new implementation because it used to be the `Bearer` string for the `Authorization` header, and V2 doesn't use that).

I can leave the members in, but they might not make sense if someone has actually created a derived class that depends on them.

java/src/com/ibm/streamsx/rest/StreamingAnalyticsServiceV1.java
```diff
@@ -0,0 +1,209 @@
...
+        return "jobId";
+    }
+
+    @Override
+    protected JsonObject getBuild(String buildId, CloseableHttpClient httpclient,
+            String authorization) throws IOException {
+        String buildURL = getBuildsUrl(httpclient) + "?build_id="
+            + URLEncoder.encode(buildId, StandardCharsets.UTF_8.name());
+        HttpGet httpget = new HttpGet(buildURL);
+        httpget.addHeader("accept", ContentType.APPLICATION_JSON.getMimeType());
```
wmarshall484: This is redundant. You already specify the `APPLICATION_JSON` in the getGsonResponse method.

ddebrunner:
> change due to behaviour change of cancelJob: it can return true / false, but also declares it can throw, and did not specify what it did for an invalid job.

Is this is another breaking change for backwards compatibility.

johnmac-ibm:
@ddebrunner Depends what you consider backward compatibility. :-) I would say undocumented behaviour in an error situation changed, so reliance on the undocumented behaviour would have been unwise.

Again, I'm easily convinced either way.

ddebrunner:
@johnmac-ibm If a method is changed to declares it throws an (non-runtime) exception then it breaks backwards compat. Now maybe that isn't what changed, but that's what I took from your description.

ddebrunner:
Merging to a feature branch for continued shared development.

johnmac-ibm:
@ddebrunner Sorry, no, the method signature did not change, it was previously declared to throw Exception (which IMHO is also a tad broad), and still does. What changed is that cancel of an invalid job ID used to return `fail` but now throws. The original documentation did not say how invalid job IDs were handled, so I don't consider this a change in documented behaviour.

PR #1298, opened 2017-11-16
add StreamingAnalyticsService::submit_job() method to Python REST API

ejpring:


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
```diff
@@ -656,13 +660,14 @@ class JobConfig(object):
         job_config.add(cfg)
         context.submit('ANALYTICS_SERVICE', topo, cfg)
     """
-    def __init__(self, job_name=None, job_group=None, preload=False, data_directory=None, tracing=None):
+    def __init__(self, job_name=None, job_group=None, preload=False, data_directory=None, submission_parameters=None, target_pe_count=None, tracing=None):
```
ddebrunner: Changing the order of parameters breaks backwards compatibility.

Target pe count was not added as a parameter as I didn't want too many parameters on the constructor, since they can easily be set as properties of the object.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1180,32 @@ def __init__(self, rest_client, credentials):
     def _get_url(self, req_name):
         return self._credentials['rest_url'] + self._credentials[req_name]
 
+    def submit_job(self, sab_file, configuration=None):
+        """Submit a compiled Streams Application Bundle (a SAB file) to run in this Streamin Analytics service, optionally with a job configuration overlay.
```
ddebrunner: Don't need to say 'compiled' do we, since a sab is always compiled.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1180,32 @@ def __init__(self, rest_client, credentials):
...
+        
+        Args:
+        sab_file(str): path to a compiled SAB file containing the application to be submitted
+        configuration(dict): job configuration overlay containing application parameters, deployment options, tracing level, etc.
```
ddebrunner: Isn't this a `JobConfig` not a `dict`?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1180,32 @@ def __init__(self, rest_client, credentials):
     def _get_url(self, req_name):
         return self._credentials['rest_url'] + self._credentials[req_name]
 
+    def submit_job(self, sab_file, configuration=None):
```
ddebrunner: Good to have an example, bonus points for showing how a client gets a `Job` object from the return.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1205,7 +1229,7 @@ def start_instance(self):
         """Start the instance for this Streaming Analytics service.
 
         Returns:
-            dict: JSON response for the instance start operation.
+            dict: JSON response from service for the instance start operation.
```
ddebrunner: What value does 'from service' provide here?

test/python/rest/TestApplication/Main.spl
```diff
@@ -0,0 +1,71 @@
+/*
+** Copyright (C) 2017  International Business Machines Corporation
+** All Rights Reserved
```
ddebrunner: Not the correct copyright - since this is ALv2 licenced code. See existing files for standard copyright.

test/python/rest/string_tests.py
```diff
@@ -0,0 +1,22 @@
+import unittest
```
ddebrunner: I'm confused, what is this test to do with this project?

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,212 @@
...
+
+    return result
+
+@unittest.skipIf(not environment_variables_set() , "Streams environment variables not set")
+class RestSubmitTests(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(self):
+
+        # temporarily compile test application locally until REST API wrapper supports remote build service
```
ddebrunner: Just to note we could use topology to build the app, rather than directly invoking sc.

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,212 @@
...
+runInterval = 30
+
+def environment_variables_set():
+
+    variables = [ 'STREAMS_INSTALL', 'STREAMING_ANALYTICS_SERVICE_NAME', 'VCAP_SERVICES' ]
+    result = True
+
+    for variable in variables:
+        if variable not in os.environ:
+            logger.error(variable + ' environment variable not set')
```
ddebrunner: It's not actually an error if the variables are not set. We run the tests just with distributed for example.

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,212 @@
...
+        logger.warning('check job ' + result['name'])
+        jobs = self.instance.get_jobs(result['name'])
+        self.assertTrue(len(jobs)>0, 'submit_job() failed, could not find job')
+
+        for job in jobs:
+            self.assertTrue(result['name']==job.name, 'submit_job() failed, could not find job')
+            self.assertTrue(job.status=='running', 'submit_job() failed, job is not running, status is ' + job.status + ', health is ' + job.health)
+            self.assertTrue(job.health=='healthy', 'submit_job() failed, job is not healthy, status is ' + job.status + ', health is ' + job.health)
+
+        for job in jobs:
```
ddebrunner: This seems to be testing `get_application_log`, just wondering why it was added, since there are existing tests for that.

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,212 @@
...
+
+        logger.warning('check job ' + result['name'])
+        jobs = self.instance.get_jobs(result['name'])
+        self.assertTrue(len(jobs)>0, 'submit_job() failed, could not find job')
+        for job in jobs:
+            self.assertTrue(result['name']==job.name, 'submit_job() failed, could not find job')
+            self.assertTrue(job.status=='running', 'submit_job() failed, job is not running, status is ' + job.status + ', health is ' + job.health)
+            self.assertTrue(job.health=='healthy', 'submit_job() failed, job is not healthy, status is ' + job.status + ', health is ' + job.health)
+            pes = job.get_pes()
+            self.assertTrue(len(pes)==1, 'job configuration failed, job has more than one PE')
```
ddebrunner: I don't think it's guaranteed that a single PE is created is it?

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,212 @@
...
+
+        jobConfigOverlay = {}
+        jobConfig._add_overlays(jobConfigOverlay)
+
+        logger.warning('submit bundle ' + os.path.basename(applicationBundle))
+        result = self.service.submit_job(applicationBundle, configuration=jobConfigOverlay)
+        self.assertFalse('status_code' in result, 'submit_job() failed, status code and error description:' + str(result))
+        self.assertTrue('name' in result, 'submit_job() failed, no job name returned, result: ' + str(result))
+
+        logger.warning('let the job run a while')
```
ddebrunner: Ideally use of the existing test framework would simplify this test, as it handles waiting for the job to become healthy, cancelling it, getting the rest connection etc. Though I'm not sure it can handle submission from a sab yet. More a note to myself to improve the test harness to support testing a bundle.

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,212 @@
...
+            #    self.assertTrue(filename=='./logs/'+job.name+'.PE_'+pe.id+'.trace', 'store_appliation_trace() failed, stored ' + filename + ' instead of ' + './logs/'+job.name+'.PE_'+pe.id+'.trace')
+
+        for job in jobs:
+            logger.warning('cancel job ' + job.name)
+            result = job.cancel()
+            self.assertTrue(result, 'cancel() failed for job ' + job.name)
+
+
+    @classmethod
+    def tearDownClass(self):
```
ddebrunner: Please don't stop the service with a test - we run multiple tests against a service.

test/python/rest/submit_tests.py
wmarshall484: Since the target_pe_count is set in the JobConfig, the fusion_scheme does get set to `manual` in the JobConfig.

ddebrunner: I think the hint is in the name "target" - not sure it is guaranteed that there will be one pe, though maybe it is if there are no exloaction constraints. Just wondering about intermittent test failures if the scheduler decides it can't fuse into a single pe.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: In the other PR I said there were two options we wanted to support:

- a JCO file/object
- a `JobConfig` object

If this just supports a JCO object then it's ok - others can be added later.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ejpring: All of the unit tests show how to get a Job object from the dictionary returned by submit_job(). I could copy some of that code into the method documentation, or just link to the unit test source. Is there a way to generate HTML links in Python documentation?

test/python/rest/string_tests.py
ejpring: This comes from the example in the README file. It didn't work, so I copied the example into this source file, fixed it, tested it, and copied it back into the README file. I could delete the source file, but think working examples are more useful than commentary examples, so I'd rather replace the example in the README file with a link to this source file.

test/python/rest/submit_tests.py
ejpring: This concerns me. Are there situations where Streams will not fuse an application into a single PE when 'fusion_scheme' is 'manual' and 'pe_count' is '1'? Several of our telco customers have deployed applications that depend upon this to meet their performance requirements.

test/python/rest/submit_tests.py
ejpring: No problem.

test/python/rest/TestApplication/Main.spl
ejpring: No problem.

PR #1299, opened 2017-11-16
 add Job::store_logs(), PE:store_console_log(), PE_store_application_trace() to Python REST API

ejpring:
common download code is in _ResourceElement::_store_file()


com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1182,6 +1230,32 @@ def __init__(self, rest_client, credentials):
     def _get_url(self, req_name):
         return self._credentials['rest_url'] + self._credentials[req_name]
 
+    def submit_job(self, sab_file, configuration=None):
+        """Submit a compiled Streams Application Bundle (a SAB file) to run in this Streamin Analytics service, optionally with a job configuration overlay.
```
wmarshall484: "Streaming"

test/python/rest/string_tests.py
```diff
@@ -0,0 +1,22 @@
+import unittest
```
wmarshall484: What is the purpose of `string_tests` in the `rest` directory? It doesn't test any rest functionality.

test/python/rest/submit_tests.py
```diff
@@ -0,0 +1,206 @@
...
+
+        logger.warning('check service ' + name)
+        instances = self.connection.get_instances()
+        if not len(instances)==1: raise unittest.SkipTest(name + ' service not found')
+        self.instance = instances[0]
+        if not self.instance.status=='running': raise unittest.SkipTest(name + ' service is not running')
+        if not self.instance.health=='healthy': raise unittest.SkipTest(name + ' service is not healthy')
+
+
+    def test_A_submit_with_defaults(self):
```
wmarshall484: Could you delete the `output` directory and the downloaded logs after the tests run? Other tests make sure that artifacts are deleted.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -404,49 +441,6 @@ class Job(_ResourceElement):
         >>> print (jobs[0].health)
         healthy
     """
-    def get_application_logs(self, path=None, prefix=None):
```
wmarshall484: The `tester.py` module has a reference to `get_application_logs`. When a test fails, it would retrieve the logs:
https://github.com/ejpring/streamsx.topology/blob/814bb4f6e470d9f2fbc5e649c084bc012df7b825/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py#L539

Could you update the reference to the new method name `store_logs`?

ddebrunner: I'm confused why are we removing `get_application_logs` ?

wmarshall484:
I am also confused about the name change from `get_application_logs`. Ideally, we should model the names after the JMX api, which also has support for downloading logs.

For example, the JMX api has a `retrieveApplicationTrace` method which returns the URL for the application trace tar file: https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.jmxapi.doc/api/com/ibm/streams/management/job/PeMXBean.html#retrieveApplicationTrace-java.lang.String-

Since our method actually downloads the tar file, I think `get_application_logs` more appropriate, but I'm curious as to why you chose `store_logs` as the new name.

ddebrunner:
Looks like this also includes `submit_job` which is in a separate PR and will need changes. Is it possible to create a PR without a dependency on submit_job, so it's truely independent.

ddebrunner:
Note we need to resolve the naming quickly, as `get_application_logs` is already in 1.8 and we will need to GA that soon.

ddebrunner:
We are also wrapping the REST api so we should be closely aligned with that, so REST api has `instance` and we use `get_instance`.

For this it has `applicationLogTrace` so maybe it should be `get_application_log_trace`.

?

test/python/rest/string_tests.py
ejpring: This comes from the example in the README file. It didn't work, so I copied the example into this source file, fixed it, tested it, and copied it back into the README file. I could delete the source file, but think working examples are more useful than commentary examples, so I'd rather replace the example in the README file with a link to this source file.

test/python/rest/submit_tests.py
ejpring: Sure, no problem.

ejpring:
Usually, get_xxx() methods return 'xxx' to the caller, but this method retrieves 'xxx' from a remote server and stores it in a local file system, so I think it should be named retrieve_xxx() or store_xxx(). The 'xxx' here is both console logs and application traces from all of the job's PEs, but I wanted a shorter name. I'd be happy with something like retrieve_application_logs().

ejpring:
Sorry, I can't make a pull request for the store/retrieve_xxx() methods that's independent of the submit_job() method because I have no way of testing them without first submitting a job.

I'll cancel the two pending pull requests and try again with your 'rest_submit_job' branch ...

ddebrunner:
> Sorry, I can't make a pull request for the store/retrieve_xxx() methods that's independent of the submit_job() method because I have no way of testing them without first submitting a job.

That's not true, we submit Python jobs all the time. @wmarshall484 wrote an existing test for `get_application_log`.

ddebrunner:
>  but I wanted a shorter name. I'd be happy with something like retrieve_application_logs().
The trouble with a shorter name (including the existing name) is that information is being lost and makes assumptions about the future that may not be true. So the call today retrieves application log and trace, if we shorten that to just log, what happens in the future if there is a REST method to just get the application log, what would it be called?

ejpring:
Maybe "retrieve_all_job_logs()" ?

PR #1301, opened 2017-11-20
 roll up accumulated submit_job() and store_xxx() updates

ejpring:


ddebrunner:
I still would like separate pull requests, since these are all separate functionality, with additional minor changes added in. With separate PRs focused on a single item, most likely some of these could have been merged already.

ddebrunner:
Please see #1330

PR #1312, opened 2017-11-23
Switch to planned public interfaces / classes for REST APIs.

johnmac-ibm:
Remove `@Deprecated` tags on the classes from 1.7, but add some comments discouraging use of `StreamingAnanlyticsConnection` (recommend `StreamingAnalyticsService` instead) and `apiKey` protected member.

Make `IStreamsConnection`, `IStreamingAnalyticsConnection`, and `StreamsRestFactory` package access.

Add a convenience factory method to `StreamingAnalyticsService` (one sample, and IBM Streams Runner for Apache Beam start with a String reference to VCAP services).

Update existing uses of previously public interfaces / classes.


johnmac-ibm:
I should mention the `FIXME` in `RESTTesterRuntime` is will be fixed in a followup patch.

PR #1313, opened 2017-11-27
Merge of Sas v2 rest changes to master.

ddebrunner:
Code has been manually tested against v2 service.


johnmac-ibm:
I have a couple more cleanup PRs coming, mostly just shuffling inside the `rest` packages.

java/src/com/ibm/streamsx/rest/StreamingAnalyticsConnectionV2.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return boolean indicating
+     *         <ul>
+     *         <li>true - if job is cancelled</li>
+     *         <li>false - if the job still exists</li>
+     *         </ul>
+     * @throws IOException
+     */
+    boolean cancelJob(String instanceId, String jobId) throws IOException {
+        if (null == jobsUrl) {
+            String restUrl = jstring(credentials, "rest_url");
```
wmarshall484: Since this is part of the V2 implementation, shouldn't this be `v2_rest_url` instead of `rest_url` ?

johnmac-ibm: Yes, I have correction for that; it missed getting updated when the credentials finalized. `StreamingAnalyticsConnectionTest` noticed that, but I'm still investigating another failure (now that I have an instance to test against). I'll put up one or more new PRs once I understand the failures.

PR #1314, opened 2017-11-27
Remove IStreamingAnalyticsConnection

johnmac-ibm:
Since only remaining use is internal, use AbstractStreamingAnalyticsConnection directly.


java/src/com/ibm/streamsx/rest/StreamsRestFactory.java
```diff
@@ -86,7 +86,7 @@ public static IStreamsConnection createStreamsConnection(String userName,
      * @return a connection to IBM Streaming Analytics service
      * @throws IOException
      */
-    public static IStreamingAnalyticsConnection createStreamingAnalyticsConnection(
+    public static AbstractStreamingAnalyticsConnection createStreamingAnalyticsConnection(
```
ddebrunner: Why are these methods public if this is an internal interface?

johnmac-ibm: There's a PR coming that will remove `StreamsRestFactory` entirely, but there's no reason for them to be public now, I'll change to package.

johnmac-ibm:
I think I messed up the merge, I will push a fix

PR #1317, opened 2017-11-27
[WIP] Add StreamingAnalyticsService.getJobId()

johnmac-ibm:
The job submission result from V2 of the Streaming Analytics Service uses a member named just `id` instead of `jobId` that was in V1.

This pull request adds `StreamingAnalyticsService.getJobId()` which takes the submission result and returns the `String` job ID based on the service version.


johnmac-ibm:
@ddebrunner WIP because I hope there is a better way to handle the change in `RemoteStreamingAnalyticsServiceStreamsContext` and also suspect there may be other places that need to be changed that I have missed related to `SubmissionResultsKeys.JOB_ID` but the only instance I found related to `DISTRIBUTED` context.

PR #1321, opened 2017-11-27
Change StreamingAnalyticsService build / submit methods to take JCO, return Job

johnmac-ibm:
Current internal users were updated. Since they expected results JSON to be added to the submission JSON, they now do that directly _only_ with the returned job ID. No other results are saved.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/StreamingAnalyticsService.java
+++ b/java/src/com/ibm/streamsx/rest/StreamingAnalyticsService.java
@@ -10,7 +10,6 @@ import static com.ibm.streamsx.topology.context.AnalyticsServiceProperties.VCAP_
 
 import java.io.File;
 import java.io.IOException;
-import java.math.BigInteger;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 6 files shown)


java/src/com/ibm/streamsx/rest/StreamingAnalyticsService.java
```diff
@@ -110,34 +109,23 @@ static StreamingAnalyticsService of(JsonElement vcapServices,
...
-     * <LI>{@code graph} - Required - JSON representation of the topology graph.</LI>
-     * </UL>
-     * <p>
-     * Results are added to the submission parameters object.
      * @param archive The application archive to build.
-     * @param submission Topology and deployment info to be submitted.
-     * @return The job id, or -1. Results from the submit will be added to the
-     * submission parameter object as a member named @{code submissionResults}.
+     * @param jco Job configuration overlay in JSON format.
+     * @param buildName A name for the build, or null.
```
ddebrunner: So as coded this is not the build name since it is modified (I think) to include a random string at the end (which is the actual build name). So is this really a build name prefix?

johnmac-ibm: Correct, I'll do a final PR on Javadoc for cleanup at some point. I'm hoping not, but we may still need some changes to allow access to hosts with self-signed certs that might need some interface / doc changes as well.

PR #1325, opened 2017-11-28
Pass window marks through map, filter, flat_map, hash_adder

ddebrunner:


ddebrunner:
Cherry-picked to v1_8

Issue #1302, opened 2017-11-20
Resolve getting the application log & trace issues.


ddebrunner:
Python REST wrapper currently has a method `get_application_logs()` on the `Job` object: [1]
Its summary is:
_Retrieves the application log and trace files of the job and saves them as a tar file to the specified path with the given name._

The REST operation being wrapped is:
- an HTTP GET on `applicationLogTrace` (which will return a stream containing the compressed logs)
- a store of returned stream as a compressed tar file.

It takes a `path` to store the tar file and `prefix` of the file name, returning the file name.

It has been proposed to change this in two ways:

1. Change the name to `store_logs` or maybe `retrieve_application_logs`
   - With summary: _Download the console logs and application traces from a job and store them locally in a compressed 'tar' archive_
2. Simplify the api to just have an optional `filename` parameter that defines the filename to use for the tar file.

We need to resolve which direction we want to take.

[1]: https://github.com/IBMStreams/streamsx.topology/blob/d2daaaecee5f1953ee9e50977767609358e8e27e/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py#L407

ddebrunner:
I think it's important to note the call is getting the application *log* and *trace*. If in the future there were calls to just get the application log or trace individually, how would we evolve to that?

If it `store_logs` implies both, is it clear that `store_log` would just be *log*?

Or maybe a `log_only` parameter, though then we have to have a `log_only` and `trace_only parameter`, or maybe it's a signature of something like: `include_log=True, include_trace=True`.

ddebrunner:
@wmarshall484 @ejpring  - Please comment.

ejpring:
I propose:

**Job::retrieve_logs_traces(filename=None)**
Retrieve console logs and application traces from all of the job's PEs and store them as a compressed 'tar' archive in the specified file. If not specified, the default filename is 'jobname.tar.gz' in the current directory, where 'jobname' is the name assigned to the job when it was submitted. The method returns the name of the file it stored.

**PE::retrieve_console_log(filename=None)**
Retrieve the PE's console log and store it as uncompressed text in the specified file. If not specified, the default filename is 'PE_id.log', where 'id' is the PE's identifying number. The method returns the name of the file it stored.

**PE::retrieve_application_trace(filename=None)**
[...]

PR #1330, opened 2017-11-29
Rename to retrieve_log_trace

ddebrunner:
Renames `get_application_logs` to `retrieve_log_trace` following the discussion in #1302.

Separate to PR #1301 as we need to minimize changes prior to a release and want to get the existing method  (`get_application_logs`) consistent with the proposed names.


ddebrunner:
Cherry-picked to v1_8

PR #1333, opened 2017-12-01
Add Domain and clean-up cancel jobs methods.

ddebrunner:
The existing cancel job required an instance identifier and checked it but in reality the instance identifier always comes from the context of an `Instance` or `Job`.

The one place the instance identifier was (really) checked was  when the cancel was executed for a distributed instance using `streamtool` which uses the defaults from the environment variables (`STREAMS_DOMAIN_ID` and `STREAMS_INSTANCE_ID`). In this case while the instance id was checked, the domain id wasn't , so it was still possible to cancel the wrong job. [By being connected to one domain and cancelling through streamtool using a different default domain).

This fixes this by using passing the domain and instance identifiers to `streamtool` which required creation of a `Domain` object.

This also deprecates `StreamsConnection.canceljob` as it cancels by job identifier with no context about the instance the job is running in, since a `StreamsConnection` can be connected to a domain with many instances.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/AbstractStreamsConnection.java
+++ b/java/src/com/ibm/streamsx/rest/AbstractStreamsConnection.java
@@ -44,9 +44,9 @@ abstract class AbstractStreamsConnection implements IStreamsConnection {
      * domainInstance will only be called for distributed where we cancel using
      * streamtool and not the context provided by the REST connection object.
      * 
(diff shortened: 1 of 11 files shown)


ddebrunner:
@johnmac-ibm

ddebrunner:
Cherry-picked to v1_8.

PR #1336, opened 2017-12-04
[Fixes #1334] Handle mixed-version credentials for Streaming Analytics Service

### LINKED ISSUE #1334: Streaming Analytics Service REST API support for credentials with mixed V1 / V2 info
The recently added REST changes assumed credentials would be either for V1 or V2 but we recently learned there will be a case where credentials could contain both:

> When I refer to v2 API I am only referring to the APIs provided by Streaming Analytics. For the case where a VM-based instance has credentials containing a `rest_url` (v1) and a `v2_rest_url` you can choose to use either interface. 
>
> Regardless of whether you use our v1 or v2 interface... if the credentials contain a `user` and `password` you *must* use Basic auth for the Streams REST API. When the credentials do not contain `user` and `password`, you must use Bearer auth with an IAM token for the Streams REST APIs.

We have two options to resolve, one is simpler but might be less flexible going forward:

1. Switch the order we check so if we see `userid` and `password` treat as V1 service and use `Basic` auth, if we don't see that but do see `v2_rest_url` use V2 service and `Bearer` auth. This would only require changing the version checking utility code.
2. Separate authentication from service and Streams REST handling, providing either a Basic or Bearer authentication strategy when we create the service / connection. This is a bit more work but I think properly separates the REST APIs from the authentication required.

johnmac-ibm:
Change the order of checks to determine Streaming Analytics Service version so that mixed-version credentials will report as V1.

Since mixed versions support both V1 and V2 service APIs but only support Basic authentication for Streams REST API, treating as V1 works for the mixed version case.

Tested with V1 and V2 services, and a V1 service with hand-crafted credentials that look like what is expected from mixed version instances.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/StreamsRestUtils.java
+++ b/java/src/com/ibm/streamsx/rest/StreamsRestUtils.java
@@ -225,6 +225,29 @@ class StreamsRestUtils {
      * Ideally, the service would return version information directly, but for
      * now key off contents we expect in the credentials.
      * <p>
+     * As described by the IBM Cloud team there are 3 cases:
+     * <ol>
+     * <li>New container-based RHEL 7.x instances use IAM authentication for both
+     * Streaming Analytics Service and Streams REST APIs, must use V2 of the
+     * Service API, and have <tt>v2_rest_url</tt> but neither <tt>userid</tt>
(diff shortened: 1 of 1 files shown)


ddebrunner:
Cherry-picked to v1_8

PR #1341, opened 2017-12-06
Add checkStatus

ddebrunner:
Add `StreamingAnalyticsService.checkStatus`.

Since the REST api is a wrapper around the Service REST api don't perform check status calls in the submit job methods, instead that's a choice for *clients* of the api.

Now topology as a client performs the check status before it submits a job.


ddebrunner:
Cherry-picked to v1_8

PR #1346, opened 2017-12-07
Eliminate extra call to V2 REST API to set up jobs / builds URLs

johnmac-ibm:
When getting the status to check the instance (which normally happens early) give the V2 implementation an opportunity to stash the jobs / builds URLs.

If there was no check, they will still call explicitly to set when required.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/AbstractStreamingAnalyticsService.java
+++ b/java/src/com/ibm/streamsx/rest/AbstractStreamingAnalyticsService.java
@@ -117,6 +117,14 @@ abstract class AbstractStreamingAnalyticsService implements StreamingAnalyticsSe
         this.authorization = authorization;
     }
 
+    /**
+     * Called on checking instance to allow implementations to update state
+     * if required.
+     */
+    protected void updateStatus(JsonObject statusResponse) {
+        // Do nothing by default, implementation classes may override
+    }
+
     @Override
     public Result<Job, JsonObject> submitJob(File bundle, JsonObject jco) throws IOException {
         final CloseableHttpClient httpClient = HttpClients.createDefault();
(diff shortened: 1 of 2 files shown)


java/src/com/ibm/streamsx/rest/AbstractStreamingAnalyticsService.java
```diff
@@ -117,6 +117,14 @@ protected void setAuthorization(String authorization) {
         this.authorization = authorization;
     }
 
+    /**
+     * Called on checking instance to allow implementations to update state
+     * if required.
+     */
+    protected void updateStatus(JsonObject statusResponse) {
```
ddebrunner: The name `updateStatus` confused me as I thought it was somehow related to checking the status, but it's really `updateState`.

However, since this is specific to v2, do we need to make it an "abstract" api in the super class? Why not just handle it completely in v2 and then it can be clearly called `fetchUrls` or similar?

Maybe it seems the name 'updateStatus' is actually coming from the v2 implementation?

johnmac-ibm: You're right that's where the name came from. I put it up in the abstract class because at the moment the `checkStatus` implementation is there and at present the only chance the V2 implementation would have to grab the response from that was to add something there.

If the goal is to save the extra REST call that was happening because the checking of the instance didn't update the URLs so `setUrls` had to, then I think either `checkStatus` has to move some of its guts to the implementation classes (eg. change `getStatusUrl` to `getStatus` that returns JSON and can extract the URLs) or it needs to pass the response back to the implementation as in this patch (though perhaps with a better name).

Or I'm missing something, which is certainly possible. :-)

johnmac-ibm:
Alternate take with overrides only in the V2 implementation.

PR #1348, opened 2017-12-08
Allow submission time parameter as topic when microservice is created with Java Application API

chanskw:
@ddebrunner this is a change that James has made to the topology toolkit to allow for topic to be specified as submission parameter, when the microservice is written in Java.

He did not have a chance to contribute this and we do not have testcases to test it.

I am wondering if you can take a look at this and see if you want to pull this in.  

One of the health toolkit samples is built on top this feature...and we can't run without it at the moment.  We are running off our own version of the topology toolkit, which is probably not great in the long run.

Please take a look.  Thanks!

### DIFF
--- a/java/src/com/ibm/streamsx/topology/spl/SPLStreams.java
+++ b/java/src/com/ibm/streamsx/topology/spl/SPLStreams.java
@@ -21,9 +21,11 @@ import com.ibm.streamsx.topology.builder.BInputPort;
 import com.ibm.streamsx.topology.builder.BOperatorInvocation;
 import com.ibm.streamsx.topology.function.BiFunction;
 import com.ibm.streamsx.topology.function.Function;
+import com.ibm.streamsx.topology.function.Supplier;
 import com.ibm.streamsx.topology.internal.core.JavaFunctional;
 import com.ibm.streamsx.topology.internal.core.JavaFunctionalOps;
 import com.ibm.streamsx.topology.internal.logic.LogicUtils;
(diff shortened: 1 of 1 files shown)


java/src/com/ibm/streamsx/topology/spl/SPLStreams.java
```diff
@@ -43,9 +45,19 @@
      */
     public static SPLStream subscribe(TopologyElement te, String topic,
             StreamSchema schema) {
+    	return subscribe(te, Value.of(topic), schema);
+    }
 
-        Map<String, Object> params = new HashMap<>();
+    public static SPLStream subscribe(TopologyElement te, Supplier<String> topic, StreamSchema schema) {
```
ddebrunner: Needs javadoc and an @since tag (1.8 if this is done quickly)

java/src/com/ibm/streamsx/topology/spl/SPLStreams.java
```diff
@@ -43,9 +45,19 @@
...
+    	return _subscribe(te, topic, schema);
+    }
 
+    private static SPLStream _subscribe(TopologyElement te, Supplier<String> topic, StreamSchema schema) {
```
ddebrunner: Not sure why an `_subscribe` was added, it's not needed the way it's coded, though it might be needed with `Object` as the type of `topic`.

ddebrunner:
Handling as a feature through this PR as it needs more development: https://github.com/ddebrunner/streamsx.topology/pull/17

PR #1352, opened 2017-12-12
Support for version 2 of the Streaming Analytics service, including IAM authorization and an update REST interface.

wmarshall484:


test/python/topology/test2_placement.py
```diff
@@ -33,22 +33,26 @@ def __call__(self, t):
...
-     h = h.map(RemoveDup())
-     
-     tester = Tester(topo)
-     tester.tuple_count(h, 2)
+    def setUp(self):
+        Tester.setup_streaming_analytics(self, force_remote_build=True)
+
+    def test_ResourceTags(self):
+        # Host tags might not be present when using a v2 service
+        if self.is_v2:
```
ddebrunner: What sets `self.is_v2` ?

test/python/rest/test_rest_bluemix.py
```diff
@@ -15,3 +16,22 @@ def setUp(self):
         Tester.setup_streaming_analytics(self, force_remote_build=True)
         self.sc = StreamingAnalyticsConnection()
         self.test_config[ConfigParams.STREAMS_CONNECTION]=self.sc
+
+        self.is_v2 = False
```
ddebrunner: Is this used?

test/python/topology/test_streaming_analytics_service.py
```diff
@@ -153,13 +154,22 @@ def test_submit_job_results(self):
...
-        self.assertIn("jobId", rc, "\"jobId\" field not in returned json dict")
-        self.assertIn("application", rc, "\"application\" field not in returned json dict")
-        self.assertIn("name", rc, "\"name\" field not in returned json dict")
-        self.assertIn("state", rc, "\"state\" field not in returned json dict")
-        self.assertIn("plan", rc, "\"plan\" field not in returned json dict")
-        self.assertIn("enabled", rc, "\"enabled\" field not in returned json dict")
-        self.assertIn("status", rc, "\"status\" field not in returned json dict")
-        self.assertIn("instanceId", rc, '"instanceId" field not in returned json dict')
 
+        if not self.is_v2:
```
ddebrunner: Could move common cases into a single block, e.g jobId, application etc.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
```diff
@@ -217,6 +218,21 @@ def setup_streaming_analytics(test, service_name=None, force_remote_build=False)
...
 
+        vcap_services = os.environ.get('VCAP_SERVICES')
+        if not vcap_services.startswith('/'):
+            #vcap services is json
+            vcap_services = json.loads(vcap_services)
+        else:
+            with open(vcap_services) as fh:
+                vcap_services = json.load(fh)
+
+        test.is_v2 = False
```
ddebrunner: This (`is_v2`) is a really generic name to write into someone's test case object. I also don't see it documented anywhere.

Is there a use case to have it exposed, or is this really only for the special case of the topology tests which need to run against both?

test/python/topology/test2_placement.py
wmarshall484: tester.setup_streaming_analytics.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -137,17 +139,105 @@ def __init__(self, username, password):
...
+        return self._bearer_token
+
+    def _refresh_authorization(self):
+        post_url = self._token_url + '?' + self._get_token_params(self._api_key)
+        res = requests.post(post_url, headers = {'Accept' : 'application/json',
+                                                 'Content-Type' : 'application/x-www-form-urlencoded'})
+        self.handle_http_errors(res)
+        res = res.json()
+
+        self._auth_exporiry_time = int(res[_IAMConstants.EXPIRATION]) - _IAMConstants.EXPIRY_PAD_MS
```
ddebrunner: Typo! Needs to be `_auth_expiry_time`

ddebrunner: Is there a mismatch of units here?
Above (line 199) `cur_time` represents the number of seconds since the epoch, but in setting the new expiry time "MS" is being used, which seems to be milliseconds.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1160,10 +1250,119 @@ def get_view_obj(_view, rc):
...
 class StreamingAnalyticsService(object):
     """Streaming Analytics service running on IBM Bluemix cloud platform.
     """
+    def __init__(self, rest_client, credentials):
+        # If IAM, create a V2 delegator, if basic http auth, create a V1 delegator
+        #
+        # Delegators are required because we need to keep StreamingAnalyticsService
+        # around for backwards compatibility, yet it also needs to work with basic
+        # and IAM authentication.
+        if 'v2_rest_url' in credentials:
```
ddebrunner: I'm not sure this is correct. `_StreamingAnalyticsServiceV2Delegator` always seems to assume a IAM rest client but I think it can be passed a user/password REST client.

I think `_StreamingAnalyticsServiceV2Delegator` always assumes IAM because it always calls `_get_authorization` which doesn't seem to exist in the non-IAM client.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Yes, cur_time needs to be converted to ms.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
wmarshall484: Yes, it's "only" for that special case. While it should be documented, it does seem that a handful of tests depend on the service version.

test/python/rest/test_rest_bluemix.py
wmarshall484: Yes, in `primitives_caller`, we need to check the service version to determine whether the resourceAllocations should be retrieved from the job.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Right, so instead, it should look like:
```python
if 'username' in credentials and 'password' in credentials:
    # use v1
else:
    # use v2
```

wmarshall484: Would you agree?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
ddebrunner: Why not just have it in topology unittests, and not the tester.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: I added a `_get_authorization` method to _StreamsRestClient:
```python
    def _get_authorization(self):
        return requests.auth._basic_auth_str(self.username, self.password)
```

test/python/topology/test_streaming_analytics_service.py
wmarshall484: Done

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: Isn't that a fixed value for v1, that could be calculated once in `__init__`?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Fixed.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
wmarshall484: Fixed.

test/python/rest/test_rest_bluemix.py
```diff
@@ -15,3 +16,22 @@ def setUp(self):
...
+        if _IAMConstants.V2_REST_URL in self.sc.credentials:
+            self.is_v2 = True
+
+    def test_service_stop_start(self):
```
ddebrunner: Does this test run when the whole suite is run? We don't want by default tests that stop & start the service.

test/python/topology/test2_job_config.py
```diff
@@ -17,10 +17,18 @@
...
+      self.is_v2 = False
+      for creds in self.test_config['topology.service.vcap']['streaming-analytics']:
+        if creds['name'] == self.test_config['topology.service.name']:
+          if 'v2_rest_url' in creds['credentials']:
+            self.is_v2 = True            
 
   def test_UnicodeJobName(self):
      """ Test unicode topo names
      """
+     if self.is_v2:
```
ddebrunner: This is not true, there's a bug in the service, which needs to be fixed. We should not be changing the test due to a service bug.

ddebrunner: Is it possible to mark it as a known failure for now, rather than a skip?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
```diff
@@ -494,7 +504,7 @@ def _distributed_test(self, config, username, password):
         if sjr['return_code'] != 0:
             _logger.error("Failed to submit job to distributed instance.")
             return False
-        return self._distributed_wait_for_result()
+        return self._distributed_wait_for_result(stc.ContextTypes.DISTRIBUTED)
```
ddebrunner: This seems a random change? Not seeing a corresponding change when the context is the service?

test/python/rest/test_rest_bluemix.py
wmarshall484: I changed the test such that it needs to be run explicitly.

PR #1353, opened 2017-12-12
Implement Java TStream.setConsistent 

ddebrunner:
The Java api existed but just threw a WIP exception.

Now the consistent annotation info is maintained in the JSON graph and used to produce a `@consistent` annotation.


ddebrunner:
backported to v1_8

PR #1365, opened 2017-12-21
[WIP] rtc39581: Add support for optional data types to the topology toolkit

nigelgodwin:
Support optional types as in wiki design for:
- Java Application API
- Python decorated operators


com.ibm.streamsx.topology/opt/python/include/splpy_general.h
```diff
@@ -788,6 +835,16 @@ class SplpyGeneral {
         }
         return pySet;
     }
+ 
+#ifdef SPL_RUNTIME_TYPE_OPTIONAL_H 
+    // SPL optional tyoe to Python optional tyoe
```
ddebrunner: It's to a Python object, not a Python optional type, since there's no such thing.

com.ibm.streamsx.topology/opt/python/include/splpy_general.h
```diff
@@ -788,6 +835,16 @@ class SplpyGeneral {
...
+    template <typename T>
+    inline PyObject * pySplValueToPyObject(const SPL::optional<T> & o) {
+        if (o.isPresent())
+             return pySplValueToPyObject(o.value());
+        return SplpyGeneral::getNone();
```
ddebrunner: This is broken because the reference count for None is never increased.

com.ibm.streamsx.topology/opt/python/packages/streamsx/spl/types.py
```diff
@@ -183,3 +183,11 @@ def rstring(value):
     Create an SPL ``rstring`` value.
     """
     return Expression('RSTRING', str(value))
+
+_null = Expression.expression("null")
+
+def valueNull():
+    """
+    Create an SPL ``null`` value.
```
ddebrunner: Need to have a returns clause in the doc string, and an `versionadded`.

java/src/com/ibm/streamsx/topology/spl/SPL.java
```diff
@@ -97,6 +98,17 @@
...
     }
     
+    /**
+     * Create an SPL value wrapper object for an SPL null value.
+     * <p>
+     * Use of this is required to construct an SPL operator parameter
+     * null value for an optional type.
+     * @return the wrapper object
+     */
+    public static Object createNullValue() {
```
ddebrunner: Just `createNull` ?

java/src/com/ibm/streamsx/topology/spl/SPLNullValue.java
```diff
@@ -0,0 +1,23 @@
+/*
+# Licensed Materials - Property of IBM
+# Copyright IBM Corp. 2015
```
ddebrunner: 2017

com.ibm.streamsx.topology/opt/python/include/splpy_setup.h
```diff
@@ -133,6 +131,11 @@ class SplpySetup {
           SplpyGeneral::loadFunction("streamsx.spl.types", "_get_timestamp_tuple"));
        SplpyGeneral::decimalClass(
           SplpyGeneral::loadFunction("decimal", "Decimal"));
+
+       PyObject *fn = SplpyGeneral::loadFunction("streamsx.spl.types", "null");
```
ddebrunner: What function is this calling, I don't see it defined anywhere?

com.ibm.streamsx.topology/opt/python/packages/streamsx/spl/types.py
```diff
@@ -183,3 +183,11 @@ def rstring(value):
     Create an SPL ``rstring`` value.
     """
     return Expression('RSTRING', str(value))
+
+_null = Expression.expression("null")
+
+def valueNull():
```
ddebrunner: Python does not use camelcase for function names, but why not simply have the field be `NULL` and no need for a function? Or have the function be `null()`.

java/src/com/ibm/streamsx/topology/spl/SPLNullValue.java
ddebrunner: though do we need to define a new class for a single function?

test/java/build.xml
```diff
@@ -20,6 +20,7 @@
   <property name="samples.src.dir" location="../../samples/java/functional/src"/>
   <property name="testtk" location="../spl/testtk"/>
   <property name="testtk.java" location="${testtk}/impl/java"/>
+  <property name="testtkopt" location="../spl/testtkopt"/>
```
ddebrunner: I think this will break the build when the version required by the toolkit is not available???

test/java/src/com/ibm/streamsx/topology/test/splpy/PythonFunctionalOperatorsTest.java
```diff
@@ -610,6 +620,180 @@ public void testReturnDict() throws Exception {
         assertEquals(-64, m4.getInt("e"));
     }
     
+    /**
```
ddebrunner: This should probably be a Python test, not Java.

com.ibm.streamsx.topology/opt/python/codegen/py_pyTupleTosplTuple.cgt
```diff
@@ -37,7 +37,13 @@ void MY_OPERATOR::fromPythonToPort<%=$oport->getIndex()%>(PyObject *pyTuple <%=$
     if (<%=$ai%> < frs) {
          // Value from the Python function
          PyObject *pyAttrValue = PyTuple_GET_ITEM(pyTuple, <%=$ai%>);
-         if (!SplpyGeneral::isNone(pyAttrValue)) {
+         if (!SplpyGeneral::isNone(pyAttrValue)
+<%
+    if (SPL::CodeGen::Type::isOptional($atype)) {
```
ddebrunner: Not sure I understand the added code.  If `pyAttrValue` is the SPL NULL then the first condition will be true so the `isSplNull` is short circuited?

com.ibm.streamsx.topology/opt/python/codegen/splpy.pm
```diff
@@ -6,6 +6,21 @@ sub splToPythonConversionCheck{
 
     my ($type) = @_;
 
+    # use eval in case optional types are not supported
+    my $isOptionalType = eval {
```
ddebrunner: Please use the `Context.getProductVersion()` to check the version.

com.ibm.streamsx.topology/opt/python/include/splpy_setup.h
ddebrunner: `SplpyGeneral::callFunction` is intended for situations like this.

com.ibm.streamsx.topology/opt/python/packages/streamsx/spl/runtime.py
```diff
@@ -41,23 +41,26 @@ def _wf():
 # may be sparse, values not set by the dictionary
 # (etc.) are set to None in the Python tuple.
 
+from streamsx.spl.types import valueNull as types
```
ddebrunner: Not sure what is going on here, would seem it's trying to call `valueNull.valueNull()` ???

com.ibm.streamsx.topology/opt/python/include/splpy_setup.h
ddebrunner: Technically this code is in the wrong location, it's in a function called setupClasses which sets up classes, but this is setting a value.

com.ibm.streamsx.topology/opt/python/include/splpy_setup.h
```diff
@@ -131,11 +140,23 @@ class SplpySetup {
...
+
+   static void setupSplNull() {
+       SplpyGIL lock;
 
-       PyObject *fn = SplpyGeneral::loadFunction("streamsx.spl.types", "null");
-       SplpyGeneral::isSplNull(
-          SplpyGeneral::pyCallObject(fn, NULL));
-       Py_DECREF(fn);
+       // Get a pointer to SPL null.
+       PyObject *null = SplpyGeneral::callFunction("streamsx.spl.types", "null", NULL, NULL);
```
ddebrunner: Maybe for readability use variable name of `splnull` or `spl_null` to avoid mentally thinking that this is `NULL` or `null` from Java.

com.ibm.streamsx.topology/opt/python/include/splpy_setup.h
```diff
@@ -131,11 +140,23 @@ class SplpySetup {
...
+
+        // Call isSplNull() passing in a pointer to SPL null which will
+        // be the first caller (as this is in setup)
+        // and thus set the local pointer to SPL null.
+        bool in = SplpyGeneral::isSplNull(null);
+        if (!in) {
+          throw SplpyGeneral::generalException("setup",
+                        "Internal error - SPL null handling");
+        }
+        Py_DECREF(null);
```
ddebrunner: Need to not decrement the reference. The function `isSplNull` holds onto the reference so it needs to have a count associated with it.

com.ibm.streamsx.topology/opt/python/codegen/py_pyTupleTosplTuple.cgt
```diff
@@ -37,7 +37,19 @@ void MY_OPERATOR::fromPythonToPort<%=$oport->getIndex()%>(PyObject *pyTuple <%=$
     if (<%=$ai%> < frs) {
          // Value from the Python function
          PyObject *pyAttrValue = PyTuple_GET_ITEM(pyTuple, <%=$ai%>);
-         if (!SplpyGeneral::isNone(pyAttrValue)) {
+<%
+    # None in Python tuple is always mapped to default, or matching input.
+    # None in Python dict was remapped temporarily to Python SPL null.
+    if (hasOptionalTypesSupport() && SPL::CodeGen::Type::isOptional($atype)) {
+         # For optional types let dict None map to default value of null.
```
ddebrunner: I found this code really hard to understand, the comment on 44 doesn't actually tell the full story, in fact it's misleading as there are no dict's here, the None in dict has already been converted to an SPL NULL in a tuple.

com.ibm.streamsx.topology/opt/python/packages/streamsx/spl/runtime.py
```diff
@@ -45,22 +47,31 @@ def _splpy_convert_tuple(attributes):
     """Create a function that converts tuples to
     be submitted as dict objects into Python tuples
     with the value by position.
+    Dict values set to Python None are reset to SPL null
+    so they can be mapped later to value not present for optional types.
     Return function handles tuple,dict,list[tuple|dict|None],None
     """
 
+    def _dict_to_tuple(tuple_):
```
ddebrunner: I think this is not ok as it modifies the caller's tuple.

PR #1370, opened 2018-01-02
Fix #1369 - Blob handling what SPL tuple passed as a Python tuple

### LINKED ISSUE #1369: error: ‘pyMvs’ was not declared in this scope
If a SPL schema contains a `blob` attribute and is passed into a Python function as a `tuple` then you get:

`error: ‘pyMvs’ was not declared in this scope`

ddebrunner:


ddebrunner:
Cherry-picked to v1_8

PR #1375, opened 2018-01-04
ClassCountCountWindow test no longer depends on current year.

wmarshall484:


ddebrunner:
Is there a way to avoid use of the current year?

This is improved but is still subject to a small chance of failure if the topology is declared at the end of the year but runs a few seconds later in the new year.

wmarshall484:
Removed dependency on current year.

ddebrunner:
Cherry-picked to v1_8

PR #1379, opened 2018-01-05
Add support for raw job config overlay settings

### LINKED ISSUE #1378: Allow JobConfig to accept raw job config overlay.
`JobConfig` does not expose the full capabilities of Job Config overlays and has the potential to always be behind as new features are added.

Add the ability to provide a `dict` object that contains "raw" job overlay information that is merged with the logical items supported by `JobConfig`, e.g. job name.

An example to set the threading model:

```
jc = streamsx.topology.context.JobConfig('SchemaPassing')
jc.raw_overlay = {'deploymentConfig': {'threadingModel': 'manual'}}
```

http://ibmstreams.github.io/streamsx.topology/doc/releases/latest/pythondoc/streamsx.topology.context.html?highlight=context#streamsx.topology.context.JobConfig

Job config overlay ref: https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.ref.doc/doc/submitjobparameters.html

ddebrunner:
Fixes #1378

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -684,7 +684,7 @@ class JobConfig(object):
         cfg = {}
         job_config = JobConfig(job_name='NewsIngester')
         job_config.add(cfg)
-        context.submit('ANALYTICS_SERVICE', topo, cfg)
+        context.submit('STREAMING_ANALYTICS_SERVICE', topo, cfg)
     """
     def __init__(self, job_name=None, job_group=None, preload=False, data_directory=None, tracing=None):
         self.job_name = job_name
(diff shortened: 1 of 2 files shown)


wmarshall484:
Since there's no error checking if the user creates an invalid job config, what happens when the user creates a config with something other than the following parameters:
```
jobName
jobGroup
submissionParameters
dataDirectory
preloadApplicationBundles
tracing
```

Does it fail? If so, where?

Granted, it's hard to do proper error checking here because if new jobconfig params are added in the future, then we'd need to update our code as well.

ddebrunner:
No error checking as the whole point is to support job config overlay options that the `JobConfig` class knows nothing about. There are many more properties supported that `JobConfig` currently supports already.

ddebrunner:
Hence the name 'raw', application is expected to provide raw, valid Job Config overlay.

PR #1380, opened 2018-01-05
Fix #1377 - Correct reference counting for timestamp conversion

### LINKED ISSUE #1377: splpy_general.h: timestampGetter reference count appears incorrect
in file splpy_general.h, lines 488, 490, call of timestampGetter:
- "tst" needs reference count decremented (after using it)
- "args" reference count was already stolen, so should not be decremented

ddebrunner:
Verified with manual test, saw memory leak with 1.8.2, none with the fix.

@nigelgodwin FYI


ddebrunner:
Cherry-picked to v1_8.

PR #1384, opened 2018-01-10
[WIP]: Optional type: add dependencies on product optional type support

nigelgodwin:
There are a number of ways of doing this. Would like to discuss the various approaches briefly.


ddebrunner:
Wrong target branch?

PR #1401, opened 2018-01-15
Ensure __exit__ is called for @spl.source operator. Fixes #1397

### LINKED ISSUE #1397: @spl.source decorated operators do not call __exit__
The `__enter__` method is called, but `__exit__` is never called.

ddebrunner:


ddebrunner:
Cherry-picked to v1_8

PR #1405, opened 2018-01-16
Attempt to initialize StreamsConnection from environment variables.

wmarshall484:
Presently, it just tries to use the VM defaults only. When running distributed unit tests in a non-VM environment, some tests will incorrectly have the username and password set to `streamsadmin` and `passw0rd`.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
@@ -186,9 +186,6 @@ class Tester(object):
         if not 'STREAMS_DOMAIN_ID' in os.environ:
             raise unittest.SkipTest("Skipped due to STREAMS_DOMAIN_ID environment variable not set")
 
-        test.username = os.getenv("STREAMS_USERNAME", "streamsadmin")
-        test.password = os.getenv("STREAMS_PASSWORD", "passw0rd")
-
         test.test_ctxtype = stc.ContextTypes.DISTRIBUTED
         test.test_config = {}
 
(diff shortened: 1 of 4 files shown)


ddebrunner:
Are these standard Streams environment variables?

test/python/topology/test2_unicode.py
```diff
@@ -76,10 +76,10 @@ def setUp(self):
         Tester.setup_distributed(self)
 
         # Get username and password
-        username = self.username
-        password = self.password
+        self.username = os.getenv("STREAMS_USERNAME", "streamsadmin")
```
ddebrunner: Any reason these are fields and not just local variables?

test/python/topology/test2_submission_result.py
```diff
@@ -12,6 +12,8 @@
 class TestSubmissionResult(unittest.TestCase):
     def setUp(self):
         Tester.setup_distributed(self)
+        self.username = os.getenv("STREAMS_USERNAME", "streamsadmin")
```
ddebrunner: Seems to be new fields added with no added use of them?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
```diff
@@ -424,7 +422,9 @@ def test(self, ctxtype, config=None, assert_on_fail=True, username=None, passwor
             config: Configuration for submission.
             assert_on_fail(bool): True to raise an assertion if the test fails, False to return the passed status.
             username(str): username for distributed tests
+                .. deprecated:: 1.8.3
```
ddebrunner: Need an indication of what a deprecated item is being replaced with.

ddebrunner:
Do we need to document these environment variables?

test/python/topology/test2_submission_result.py
wmarshall484: They are used on line 28 of the same file. These fields were previously set in `Tester`, but are now set in the `setUp` method.

wmarshall484:
Added doc in commit [474ba4a](https://github.com/IBMStreams/streamsx.topology/pull/1405/commits/474ba4a184f1502846433a8271d37dd1dfb66b6d)

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
wmarshall484: Added doc indicating environment variables should be used.

PR #1409, opened 2018-01-17
Fix 1408: Indicate progress if any condition increases its sequence

ddebrunner:


ddebrunner:
Cherry-picked to v1_8

PR #1430, opened 2018-01-23
[WIP]: remove isSplNull; add support for optional types in Python app…

nigelgodwin:
…lication API

Needs fix in SPL (optional type branch) for all tests to pass.
Not all tests have been run (python2, build and test with Streams with no optional type support).


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/graph.py
```diff
@@ -225,7 +225,11 @@ def addOutputPort(self, oWidth=None, name=None, inputPort=None, schema= CommonSc
         return oport
 
     def setParameters(self, params):
+        import streamsx.spl.op
         for param in params:
+            if params[param] == None:
```
ddebrunner: Need to use `is None` not `==`

PR #1439, opened 2018-01-25
Fix code coverage reporting.

wmarshall484:
When running unittests, the jacoco code coverage would report zero coverage. This happened because the ant script was looking for `jacoco.exec` in the `unittest` directory, when it was generated at the top level.

Added a fix which looks for `jacoco.exec` at the top level.


ddebrunner:
Wonder why it changed ...

wmarshall484:
It had to do with the version of jacoco I was using. It seems version `0.7.9` places `jacoco.exec` in at the top level by default, which is different from what version `0.7.5` does.

I made a change which explicitly states `jacoco.exec` should be placed in the `unittests` directory.

Issue #1429, opened 2018-01-23
Add parallel info to Operator in Java REST API


johnmac-ibm:
The REST information for operator includes several members that are only returned if the operator is in a parallel region: `channel`, `channels` (and its nested properites), and `logicalName`. See https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.0/com.ibm.streams.restapi.doc/doc/restapis-operator.html

I'm specifically interested in `logicalName`, and happy to do the pull request but wanted to ask a couple questions first:

1. Since `logicalName` is all I need now, and since `channels` in particular could bulk up the object, should I just stick with it or do them all, and
2. Should I mimic the REST API so the getters return `null` if the property is not present, leaving that as the only indicator of a parallel region, or make them a bit more user-friendly, eg. add a `isInParallelRegion()` and have `getLogicalName()` return the bare operator name if it is not in a parallel region?

ddebrunner:
I think channel should default to the value returned by the SPL/Java/C++ functions when not in a parallel region (`-1` I think, but need to verify).
Logical name should always return the logical name (which as you say is the physical name when not in a parallel region).
If we support channel, then we need max channels as well, is that a separate property or part of the `channels` object?

johnmac-ibm:
It doesn't appear to be a separate property, but it may be implied by the `channels` array (eg. size of), though I haven't yet looked closely at what it returns in practice, beyond what the doc says:

> **channels**
> An array of objects that provide the following information about the channels of the parallel regions used to route tuple data for the operator.
> - **indexWithinRegion**
>        A number that identifies the channel index within the parallel region.
> - **logicalOperatorName**
>        A string that identifies the logical name of the parallelized operator introducing the region.

johnmac-ibm:
After playing around with it a bit, neither `logicalName` nor `channels` is not quite what I expected.

Both `name` and `logicalName` include what appear to be default names for the parallel region as a prefix to the invocation name, with the latter just lacking any index, eg. for an operator assigned `invocationName("Foo")` in a nested parallel instance we might see:

`"name": "PARALLEL[1].PARALLEL_2[3].Foo"`
`"logicalName": "PARALLEL.PARALLEL_2.Foo"`

I was a bit surprised to see the `PARALLEL.PARALLEL_2` portion, but I understand it.

The `channel` for that operator is 3, and the `channels` array has an element for each nested region:

```
"channels" : [
[...]

PR #1444, opened 2018-01-26
[Issue #1429] Add getLocialName() to Java REST Operator

johnmac-ibm:
Returns the logical name from the REST API if the operator is in a parallel region, or the plain name otherwise.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/Operator.java
+++ b/java/src/com/ibm/streamsx/rest/Operator.java
@@ -33,6 +33,8 @@ public class Operator extends Element {
     @Expose
     private String job;
     @Expose
+    private String logicalName;
+    @Expose
     private String metrics;
     @Expose
     private String name;
@@ -62,6 +64,16 @@ public class Operator extends Element {
             opList = Collections.<Operator> emptyList();
         }
         return opList;
+    }
+
+    /**
+     * The logical name of this operator.
+     * 
+     * @return the logical name of the operator, which is just the name if the
+     * operator is not part of a parallel region.
+     */
+    public String getLogicalName() {
+        return logicalName == null ? name : logicalName;
     }
 
     /**
--- a/java/src/com/ibm/streamsx/rest/Operator.java
+++ b/java/src/com/ibm/streamsx/rest/Operator.java
@@ -71,6 +71,7 @@ public class Operator extends Element {
      * 
      * @return the logical name of the operator, which is just the name if the
      * operator is not part of a parallel region.
+     * @since 1.9
      */
     public String getLogicalName() {
         return logicalName == null ? name : logicalName;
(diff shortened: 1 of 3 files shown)


java/src/com/ibm/streamsx/rest/Operator.java
```diff
@@ -64,6 +66,16 @@
         return opList;
     }
 
+    /**
+     * The logical name of this operator.
+     * 
+     * @return the logical name of the operator, which is just the name if the
+     * operator is not part of a parallel region.
+     */
+    public String getLogicalName() {
```
ddebrunner: Can you add an `@since 1.9`.

test/java/src/com/ibm/streamsx/rest/test/ParallelOperatorsTest.java
```diff
@@ -0,0 +1,107 @@
...
+            // avoid clashes with sub-class tests
+            Topology topology = new Topology(getClass().getSimpleName(), 
+                    "JobForRESTApiTest");
+
+            TStream<Integer> source = topology.periodicSource(() -> (int) (Math.random() * 5000 + 1), 200, TimeUnit.MILLISECONDS);
+            source.invocationName("IntegerPeriodicMultiSource");
+            TStream<Integer> parallelSource = source.parallel(3);
+            TStream<Integer> sourceDouble = parallelSource.map(StreamsConnectionTest.doubleNumber());
+            sourceDouble.invocationName("Outer");
+            TStream<Integer> encore = sourceDouble.parallel(2);
```
wmarshall484: Interesting to see that this worked, since technically we don't officially support nested parallelism in the API. We don't preclude it, but it hasn't been tested.

test/java/src/com/ibm/streamsx/rest/test/ParallelOperatorsTest.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Test
+    public void testParallelOperators() throws Exception {
+        List<Operator> operators = helper.job.getOperators();
+        Iterator<Operator> curr = operators.iterator();
+
+        assertTrue("Missing source operator", curr.hasNext());
+        Operator op = curr.next();
+        assertEquals("IntegerPeriodicMultiSource", op.getLogicalName());
```
wmarshall484: Hm, does the REST API guarantee that operators will be returned in a particular order?

wmarshall484: It seems to be the case, given that the test passes, but I'm wondering if that's something we should count on for other tests.

johnmac-ibm: That surprised me too when I first looked at the existing tests, but it does. From the [Operator](https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.0/com.ibm.streams.restapi.doc/doc/restapis-operator.html) docs:

> ***name***
> A string that identifies the operator name. By default, when you retrieve a collection of operator resources, they are sorted according to this property value.

wmarshall484: So _that's_ the reasoning behind `ZIntegerTransformInteger`.

PR #1447, opened 2018-01-26
Fix #1446 - Timestamp __new__ invocation that works on 2.7/3.5

### LINKED ISSUE #1446: Timestamp class broken on Python 2.7.
Error stating `super()` takes at least one parameter.

ddebrunner:


ddebrunner:
back ported to v1_8

PR #1449, opened 2018-01-27
Convert token expiry time to milliseconds.

wmarshall484:
The token expiry time is not being converted to milliseconds. As such, a new IAM token was retrieved on every REST call.

This PR converts the expiry time to MS.


ddebrunner:
back-ported to v1_8

Issue #1442, opened 2018-01-25
Support tester on Python 2.7.


ddebrunner:
Requires that the `streamsx.ec` extension module is loaded on Python 2.7.

ddebrunner:
Using Anaconda with a 2.7 Python install this is how I created a distributed instance to test with:
```
streamtool mkinstance
streamtool setproperty --application-ev PYTHONHOME=$HOME/anaconda2
streamtool setproperty --application-ev CRYPTOGRAPHY_ALLOW_OPENSSL_100=1
```

ddebrunner:
Done.

PR #1459, opened 2018-02-06
Support streamsx.ec and tester with Python 2.7

ddebrunner:
Passes all tests except one in `test/python/topology` when running in standalone.

Failing test is that `StreamSchema.style` returns a future `dict` rather than `dict`. It should return `dict` since the Python object used to represent a stream tuple is a raw `dict` as it is created using the C api.

Initial set of work for #1442


ddebrunner:
Now also works for distributed for topology tests.

ddebrunner:
Passes REST tests.

ddebrunner:
Fixed `test_schemas.py`.

All relevant tests pass under Python 2.7.

PR #1461, opened 2018-02-06
[WIP]Update the colocation keys of all operators to their final resolved tag.

wmarshall484:


java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
```diff
@@ -309,6 +309,13 @@ void resolveColocationTags() {
             placement.addProperty(PLACEMENT_COLOCATE_KEY, singleTag);
         });
         
+        // Update the colocation keys of all operators to their final resolved tag.
+        operators(graph, op -> {
+        	JsonObject placement = object(op, CONFIG, PLACEMENT);
```
ddebrunner: It's not guaranteed that an operator has a placement value, see the operators "loop" above this.

PR #1475, opened 2018-02-09
Fix #1471 - Add remote build submission metrics such as upload time

### LINKED ISSUE #1471: Add metrics for remote build service.
In the submission response when submitting to Streaming Analytics, seems we could include some metrics to help any future diagnostics:

- Size of the build archive
- Upload time
- Time wating for build, broken into states, e.g. timeQueued, timeBuilding, etc.
- Time to submit

Adding something like this to the returned JSON object:
```
"metrics": {
   "codeArchiveSize": 234433,
   "uploadTime": 23.3,
   "queueTime": 0,
   "buildTime": 12.1,
   "submissionTime": 3.0
}
```

ddebrunner:
See #1471

### DIFF
--- a/java/src/com/ibm/streamsx/rest/AbstractStreamingAnalyticsService.java
+++ b/java/src/com/ibm/streamsx/rest/AbstractStreamingAnalyticsService.java
@@ -179,7 +179,7 @@ abstract class AbstractStreamingAnalyticsService implements StreamingAnalyticsSe
             String buildName) throws IOException {
         
         JsonObject metrics = new JsonObject();
-        metrics.addProperty("codeArchiveSize", archive.length());
+        metrics.addProperty("buildArchiveSize", archive.length());
            	
         CloseableHttpClient httpclient = HttpClients.createDefault();
         try {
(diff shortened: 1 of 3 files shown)


java/src/com/ibm/streamsx/rest/AbstractStreamingAnalyticsService.java
```diff
@@ -170,10 +170,17 @@ protected void setAuthorization(String authorization) {
...
+        return "buildState_" + state + "Time_ms";
+    }
+    
     @Override
     public Result<Job, JsonObject> buildAndSubmitJob(File archive, JsonObject jco,
             String buildName) throws IOException {
-    	
+        
+        JsonObject metrics = new JsonObject();
+        metrics.addProperty("buildArchiveSize", archive.length());
```
wmarshall484: It might be nice to have an interface with the properties of the metrics object so we can keep track of what gets set where.

ddebrunner:
Moved keys to an existing interface.

PR #1482, opened 2018-02-12
Test passes under v2 WAMS service

ddebrunner:


ddebrunner:
Cherry-picked to v1_8

PR #1490, opened 2018-02-17
Support decorated operators with Python 2.7

ddebrunner:
Still need to verify it passes all the decorated operator tests with 2.7.


ddebrunner:
Tests passing on Python 2.7 & 3.5.

PR #1495, opened 2018-02-21
Create GCompositeDef and GCompositeDefImpl.

wmarshall484:
* GCompositeDef and GCompositeDefImpl created.
* Where possible, used a gcomposite instead of the raw json graph to benefit from future quick graph searching.


ddebrunner:
I have the feeling this will evolve into a different interface, given the intention of the `GCompositeDef` interface to represent an SPL composite definition, but **its current use and description** as an item that represents the complete graph.

Almost as this should be `GMainCompositeDef` with that extending a new `GCompositeDef`.

wmarshall484:
Updated GCompositeDef javadoc to state:
> The GComposite represents the definition of an SPL composite.

PR #1496, opened 2018-02-21
Use V1/Basic when given a mixed credentials object

wmarshall484:
If given a "mixed" credentials object with `v2_rest_url`, `username`, and `password` fields, then use the V1 Streaming Analytics service REST API using HTTP basic authentication.


wmarshall484:
username -> userid committed

Issue #1487, opened 2018-02-15
Operator entities in the Streams REST API have no host url when using the container-based Streaming Analytics service


wmarshall484:
As in, the following code will yield None using the WAMs beta Streaming Analytics service:

```python
# Assuming a running job in a Streaming Analytics service
from streamsx import rest
sc = StreamingAnalyticsService()
job = sc.get_instances()[0].get_jobs()[0]
op = job.get_operators()[0]
print(op.get_host())
```
Yields `None`

We can verify this is the case by inspecting the json of the Operator entity:
```python
print(op)

{'json_rep' : {
                   ...
                   'host' : None,
                   ...
                   },
'rest_client' : "...",
'rest_self'    : "..."
} 
```

As a result of this, the `test_rest_bluemix.TestRestFeaturesBluemix.test_basic_calls` rest test fails when verifying that an operator can retrieve its host.

ddebrunner:
@wmarshall484 Any update?

ddebrunner:
Just noticed for PE.get_host() we already say this:

_Get resource this processing element is currently executing in. If the processing element is running on an externally managed resource `None` is returned._

With WAMS the resources are externally managed.

ddebrunner:
Also already documented for operator, so it's a test issue.

PR #1505, opened 2018-02-23
Extract retrieve file logic

ddebrunner:
To allow support of PE based retrieve log/trace calls.

Basically a step towards "pulling" (actually reimplementing) some of the changes from the `rest_submit_job` branch.


ddebrunner:
Passed REST tests with temp work-around for #1487

Issue #1463, opened 2018-02-07
TopologyTester should support to always collect logs


markheger:
Currently the job logs are collected only if test fails. It would be good to have an option, that the Tester always collects the job logs before cancelling the job. Maybe a new optional parameter could be added to the `Tester::test` method.

`always_collect_logs=True`

ddebrunner:
Fixed by #1516

PR #1516, opened 2018-02-27
Add support for always collecting test logs.

wmarshall484:
By passing a parameter to `Tester.test` the logs will always be retrieved when running tests.


wmarshall484:
fixes #1463

PR #1517, opened 2018-02-27
Fix #1487 - Allow for get_host returning None for WAMS service

### LINKED ISSUE #1487: Operator entities in the Streams REST API have no host url when using the container-based Streaming Analytics service

ddebrunner:


ddebrunner:
All tests under test/python/rest pass for "legacy", VM, and WAMS beta services.

PR #1519, opened 2018-02-28
Add submit_job

ddebrunner:
Adds `StreamingAnalyticsService.submit_job`  from a `sab` file.

Added for V1 & V2 services, tested on RH6/7 using the added test against legacy, VM & WAMS services.

Obtians the V2 jobs  URL the "standard" way through the get status call.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
@@ -1339,13 +1339,13 @@ class StreamingAnalyticsService(object):
         else:
             self._delegator = _StreamingAnalyticsServiceV1Delegator(rest_client, credentials)
 
-    def submit_job(self, sab_file, job_config=None):
+    def submit_job(self, bundle, job_config=None):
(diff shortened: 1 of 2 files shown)


com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1387,14 +1403,41 @@ def __init__(self, rest_client, credentials):
         self.rest_client = rest_client
         self._credentials = credentials
         self._v2_rest_url = self._credentials[_IAMConstants.V2_REST_URL]
+        self._jobs_url = None
+
+    def _get_jobs_url(self):
+        """Get & save jobs URL from the status call."""
+        if self._jobs_url is None:
+            self.get_instance_status()
```
wmarshall484: I was confused when I read this initially since`get_instance_status`now has the side effect of getting the jobs url. Maybe there could be something like a `seturls` method which gets called in the delegator constructor.

ddebrunner: I'm trying to avoid an extra URL request to get URLs that might never be used.

PR #1529, opened 2018-03-02
Fix 1526 - Document user/password modes and ensure submit job uses ST…

ddebrunner:
…REAMS_USERNAME if set


ddebrunner:
I verified this fix allows topology submission using a different Streams user name to the operator system user.

The Streams domain was using an LDAP server[1] and `STREAMS_USERNAME` was set to an LDAP user `tesla`.

`tesla` had been set up for no authentication with streamtool using `streamtool genkey -U tesla`

[1] https://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/

PR #1556, opened 2018-03-16
Support for parallel sources and parallel regions with multiple inputs

wmarshall484:
Passes all parallel tests. Currently running the full test suite.

### DIFF
--- a/java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
@@ -211,37 +211,58 @@ class OperatorGenerator {
 
     private void parallelAnnotation(JsonObject op, StringBuilder sb) {
         boolean parallel = jboolean(op, "parallelOperator");
-
+        
         if (parallel) {
+            boolean partitioned = jboolean(op, "partitioned");
+            JsonObject parallelInfo = op.get("parallelInfo").getAsJsonObject();
+            
             sb.append("@parallel(width=");
-            JsonElement width = op.get("width");
+            JsonElement width = parallelInfo.get(OpProperties.WIDTH);
             if (width.isJsonPrimitive()) {
                 sb.append(width.getAsString());
             } else {
                 splValueSupportingSubmission(width.getAsJsonObject(), sb);
             }
-            String parallelInputPortName = jstring(op, "parallelInputPortName");
-            boolean partitioned = jboolean(op, "partitioned");
+      
             if (partitioned) {
+                sb.append(", partitionBy=[");
+                JsonArray partitionedPorts = array(parallelInfo, "partitionedPorts");
+                for(int i = 0; i < partitionedPorts.size(); i++){
+                    JsonObject partitionedPort = partitionedPorts.get(i).getAsJsonObject();
+                    
+                    if(i>0)
+                        sb.append(", ");
+                    
+                    sb.append("{port=");
+                    sb.append(getSPLCompatibleName(GsonUtilities.jstring(partitionedPort, "name")));
+                    sb.append(", attributes=[");
+                    JsonArray partitionKeys = partitionedPort.get("partitionedKeys").getAsJsonArray();
+                    for (int j = 0; j < partitionKeys.size(); j++) {
+                        if (j != 0)
+                            sb.append(", ");
+                        sb.append(partitionKeys.get(j).getAsString());
+                    }
+                    sb.append("]}");
+                }
                 
-                JsonArray partitionKeys = op.get("partitionedKeys").getAsJsonArray();
-
-                parallelInputPortName = getSPLCompatibleName(parallelInputPortName);
-                sb.append(", partitionBy=[{port=");
-                sb.append(parallelInputPortName);
-                sb.append(", attributes=[");
-                for (int i = 0; i < partitionKeys.size(); i++) {
+                sb.append("]"); 
+            }
+            
+            JsonArray broadcastPorts = parallelInfo.get("broadcastPorts").getAsJsonArray();
+            if(broadcastPorts.size() > 0){
+                sb.append(", broadcast=[");
+                for(int i = 0; i < broadcastPorts.size(); i++){
                     if (i != 0)
                         sb.append(", ");
-                    sb.append(partitionKeys.get(i).getAsString());
+                    sb.append(getSPLCompatibleName(broadcastPorts.get(i).getAsString()));
                 }
-                sb.append("]}]");
-            } else if ("BROADCAST".equals(jstring(op, PortProperties.ROUTING))) {
-                sb.append(", broadcast=[");
-                sb.append(parallelInputPortName);
                 sb.append("]");
+                
             }
-            sb.append(")\n");
+            
+       
+            sb.append(")");
+            sb.append("\n");
         }
     }
 
--- a/java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
@@ -364,7 +364,6 @@ class OperatorGenerator {
     }
 
     static void operatorNameAndKind(JsonObject op, StringBuilder sb, boolean singlePortSingleName) {
-
         if (!singlePortSingleName) {
             String name = jstring(op, "name");
             name = getSPLCompatibleName(name);
--- a/java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
@@ -219,6 +219,19 @@ class PEPlacement {
 
         Set<JsonObject> llStartChildren = getDownstream(llStart, graph);
         
+        // Determine if the region has already been tagged, which would happen if there are
+        // multiple starts to the low latency region.
+        Boolean[] isAlreadyTagged = {false};
+        llStartChildren.forEach(oper -> {
+            JsonObject placement = object(oper, CONFIG, PLACEMENT);
+            if(placement!=null && jstring(placement, PLACEMENT_LOW_LATENCY_REGION_ID) != null)
+                isAlreadyTagged[0]=true;
+                
+        });
+        
+        if(isAlreadyTagged[0])
+            return;
+        
         Set<BVirtualMarker> boundaries = EnumSet.of(LOW_LATENCY, END_LOW_LATENCY);
 
         GraphUtilities.visitOnce(llStartChildren, boundaries, graph,
--- a/java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
@@ -29,6 +29,7 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.google.gson.JsonObject;
 import com.ibm.streamsx.topology.builder.BVirtualMarker;
@@ -221,15 +222,15 @@ class PEPlacement {
         
         // Determine if the region has already been tagged, which would happen if there are
         // multiple starts to the low latency region.
-        Boolean[] isAlreadyTagged = {false};
+        AtomicBoolean isAlreadyTagged = new AtomicBoolean(false);
         llStartChildren.forEach(oper -> {
             JsonObject placement = object(oper, CONFIG, PLACEMENT);
             if(placement!=null && jstring(placement, PLACEMENT_LOW_LATENCY_REGION_ID) != null)
-                isAlreadyTagged[0]=true;
+                isAlreadyTagged.set(true);;
                 
         });
         
-        if(isAlreadyTagged[0])
+        if(isAlreadyTagged.get())
             return;
         
         Set<BVirtualMarker> boundaries = EnumSet.of(LOW_LATENCY, END_LOW_LATENCY);
--- a/java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
+++ b/java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
@@ -28,6 +28,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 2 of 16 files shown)


java/src/com/ibm/streamsx/topology/generator/spl/PEPlacement.java
```diff
@@ -219,6 +219,19 @@ private void assignLowLatency(JsonObject llStart) {
 
         Set<JsonObject> llStartChildren = getDownstream(llStart, graph);
         
+        // Determine if the region has already been tagged, which would happen if there are
+        // multiple starts to the low latency region.
+        Boolean[] isAlreadyTagged = {false};
```
ddebrunner: Normal approach is to use an `AtomicBoolean`.

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
```diff
@@ -211,37 +211,58 @@ private void viewAnnotation(JsonObject op, StringBuilder sb) {
...
-                parallelInputPortName = getSPLCompatibleName(parallelInputPortName);
-                sb.append(", partitionBy=[{port=");
-                sb.append(parallelInputPortName);
-                sb.append(", attributes=[");
-                for (int i = 0; i < partitionKeys.size(); i++) {
+                sb.append("]"); 
+            }
+            
+            JsonArray broadcastPorts = parallelInfo.get("broadcastPorts").getAsJsonArray();
+            if(broadcastPorts.size() > 0){
```
ddebrunner: Minor nit - use of `!broadcastPorts.isEmpty()` method is clearer.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -57,24 +61,35 @@
     private int targetRelease;
     @SuppressWarnings("unused")
     private int targetMod;
+    
+    // List of physical operator fields indicating that the operator
+    // is the start of a region
+    List<String> compOperatorStarts = new ArrayList<>();
```
ddebrunner: Why are these package protected?

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -83,6 +98,406 @@ public String generateSPL(JsonObject graph) throws IOException {
         return sb.toString();
     }
     
+    private void separateIntoComposites(JsonObject graph){
+
+        compOperatorStarts.add(OpProperties.PARALLEL);
+        compStarts.add(BVirtualMarker.PARALLEL.kind());
+        compEnds.add(BVirtualMarker.END_PARALLEL.kind());
+        
+//        compOperatorStarts.add(null);
```
ddebrunner: Why the commented out code?

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -166,18 +581,40 @@ void generateComposite(JsonObject graphConfig, JsonObject graph,
         compBuilder.append("composite ");
 
         compBuilder.append(kind);
-        if (kind.startsWith("__parallel_")) {
-            String iput = jstring(graph, "inputName");
-            String oput = jstring(graph, "outputName");
-
-            iput = getSPLCompatibleName(iput);
-            compBuilder.append("(input " + iput);
+        if (kind.startsWith("Composite")) {
```
ddebrunner: Hmm, should we use a name here that is not a valid operator kind name? Ie. a customer could in theory have a valid primitive or SPL composite called 'Composite'.

Maybe use '$Composite' and should it be in a static field as a constant?

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -83,6 +98,406 @@ public String generateSPL(JsonObject graph) throws IOException {
...
+                // Fix the naming of the operators in the composite to read from the composite input ports
+                // and write to the composite output ports
+                fixCompositeInputNaming(graph, startsEndsAndOperators, compDefinition);
+                fixCompositeOutputNaming(graph, startsEndsAndOperators, compDefinition, compInvocation);
+                
+                // Add the invocation to the graph
+                array(graph, "operators").add(compInvocation);
+
+                // Remove starts, ends, and composite operators from the graph
+                for(int j = 0; j < startsEndsAndOperators.size(); j++){
```
ddebrunner: Kind of confusing have both style loops when they could be consistent.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -83,6 +98,406 @@ public String generateSPL(JsonObject graph) throws IOException {
...
+                partitionInfo.add(PortProperties.PARTITION_KEYS, outputPort.get(PortProperties.PARTITION_KEYS));
+                partitionedPorts.add(partitionInfo);
+                
+                // There is at least one partitioned port, therefore it is partitioned.
+                compositeInvocation.addProperty("partitioned", true);
+            }      
+                
+        } 
+        
+        compositeInvocation.add("parallelInfo", parallelInfo);
```
ddebrunner: Can `parallelOperator` ever be false if there is a `parallelInfo`. 

Maybe later cleanup, but is the presence of `parallelInfo` sufficient to indicate the op is parallel.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
ddebrunner: Hmm, we need an identifier here, but we also need to ensure we don't clash with an user provided identifier.

ddebrunner: Might - delay this and fix in a subsequent PR.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
ddebrunner: OK - it's a TODO, useful to leave TODO comments for the review and as a reminder.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -83,6 +98,406 @@ public String generateSPL(JsonObject graph) throws IOException {
...
+        return compositeInvocation;
+    }
+
+    private List<List<JsonObject> > findCompositeOpsOfAType(JsonObject graph, String startKind, String endKind, String opStartParam){
+        
+        for(JsonElement jePotentialStart : graph.getAsJsonArray("operators")){
+            JsonObject potentialStart = jePotentialStart.getAsJsonObject();
+            
+            // We've found a potential start to a composite. See if the composite doesn't contain another composite.   
+            if(kind(potentialStart).equals(startKind) ||
```
ddebrunner: Need comments as to why a config section makes an operator a potential start.

Especially if most operators are going to have config sections.

Should instead a potential start be identified by a function and that what gets evaluated against the operator. Then parallel would have something like:
`kind == $Parallel$ of op config has parallel info`
while low latency would have
`kind == $LowLatency`

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -83,6 +98,406 @@ public String generateSPL(JsonObject graph) throws IOException {
...
+//        compEnds.add(BVirtualMarker.END_LOW_LATENCY.kind());
+        
+
+        // Find composites until there are no more to find.
+        while(createComposite(graph)){
+
+        }     
+    }
+    
+    /**
```
ddebrunner: I think this method is the core of the system so it should have an overview (as much desscription as possible) how you are finding and creating composites, so we can understand the algorithm later when it has gone out of our current memory.

I think renaming this method would help, because it's not creating a composite I think it's finding and creating a single composite, the one that is the first one discovered at the most nested level. Sort of `findAndCreateMostNestedComposite`.

I saw this as the methods it calls, e.g.  `findCompositeOpsOfAType` have no descriptions of what they do, and possibly the names are misleading. I think  `findCompositeOpsOfAType` is finding a single composite and its operators, not finding multiple composite operators, as its name could be interpreted as.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -83,6 +98,406 @@ public String generateSPL(JsonObject graph) throws IOException {
...
+     * When we create a composite, operators need to create connections with the composite's input port.
+     * @param graph
+     * @param startsEndsAndOperators
+     * @param opDefinition
+     */
+    private void fixCompositeInputNaming(JsonObject graph, List<List<JsonObject>> startsEndsAndOperators,
+            JsonObject opDefinition) {   
+        // For each start
+        // We iterate like this because we need to also index into the operatorDefinition's inputNames list.
+        for(int i = 0; i <  startsEndsAndOperators.get(0).size(); i++){
```
ddebrunner: Modern loops are much cleaner:
```
for(int i = 0; i <  startsEndsAndOperators.get(0).size(); i++){
    JsonObject start = startsEndsAndOperators.get(0).get(i);
```

can be:
```
for (JsonObject start : startsEndsAndOperators.get(0)) {

```

wmarshall484: But I need the counter to create a unique name for the input port. I could have a separate counter variable as well as doing the new style loop, but it seemed easier to just use a traditional loop.

ddebrunner: Which line of code is using `i` ? I just can't see it.

ddebrunner:
My guess the reason the travis/pure build is failing is the change from:
```
 else if ("BROADCAST".equals(jstring(op, PortProperties.ROUTING)))
```

to
```
PortProperties.ROUTING).equals(Routing.BROADCAST.toString())
```

Bringing TStream is bringing in "impure" code.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
wmarshall484: You uncovered a bug.
```
connections.set(j, GsonUtilities.array(opDefinition, "inputNames").get(j));
```
should be 
```
connections.set(j, GsonUtilities.array(opDefinition, "inputNames").get(i));
```

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
```diff
@@ -78,6 +78,8 @@ public String generateSPL(JsonObject graph) throws IOException {
         stvHelper = new SubmissionTimeValue(graph);
         new Preprocessor(this, graph).preprocess();
         
+        System.out.println(graph);
```
ddebrunner: In case you don't notice this, debugging printlns left in.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
wmarshall484: We don't need to keep the info in the kind name. I created a boolean field called `generated`, which is `true` if the composite was generated during SPL generation.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
wmarshall484: I like `findAndCreateMostNestedComposite`. I'm sure that was was an off-the-cuff name, but it's accurate.

java/src/com/ibm/streamsx/topology/generator/spl/OperatorGenerator.java
wmarshall484: JsonArray doesn't have an `isEmpty` method.

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
ddebrunner: Can we just clarify the terms in the description. It's understandable but only with prior knowledge of what is going on.

The issue is with "composite", it's used to mean multiple items.

Eg. _composites are non-overlapping, contiguous subsections of graph_

Well no, the graph has region of operator invocations that this algorithm extracts into a _composite definition_ and replaces the region with an _composite invocation_ of that newly defined composite.

_composites are non-overlapping, contiguous subsections of graph_

**regions** are non-overlapping, contiguous subsections of graph

_an "innermost" composite is a composite that does not contain another composite_

I'd say it's is actually: an innermost region is a region that does not contain another region. 

For a new reader the phrase "not contain another composite" might be taken as the region cannot contain an invocation of a composite.

Basically, anywhere composite it used it needs to be replaced with one of:
 * region
 * composite definition
 * composite invocation

java/src/com/ibm/streamsx/topology/generator/spl/SPLGenerator.java
wmarshall484: It's not just having a `config` section. It's both having a `config` section and the `config` containing the parameter indicating it is a `start`.

Since not all operators have configs, we first need to check if the config exists before accessing it in order to avoid a NPE.

```
op.has("config") && jboolean(object(op, "config"), opStartParam)
```

I added two functions: one which perform the logic of determining whether an operator is a (i.e., any) start op, and another which determines whether the operator is a start op for a particular type of region.

test/java/src/com/ibm/streamsx/topology/test/api/CompGenerationTest.java
```diff
@@ -75,6 +77,7 @@ public void nestedUDPWithBroadcast() throws Exception {
     @Test
     public void multipleInputPorts() throws Exception {
         assumeTrue(SC_OK);
+        assumeTrue(!isMainRun());
```
ddebrunner: Is this required? SC_OK (the check above) will rule out embedded?

wmarshall484: It didn't work without it, yes.

wmarshall484: When running `unittest.quick` it would still try to execute the test and fail on its condition.

test/java/src/com/ibm/streamsx/topology/test/api/ParallelTest.java
```diff
@@ -729,6 +729,7 @@ public void testUnionUnparallel() throws Exception {
     
     @Test
     public void testBroadcast() throws Exception {
+        assumeTrue(!isMainRun());
```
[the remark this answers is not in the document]
wmarshall484: It's not. Replaced with `checkUDPSupported`

wmarshall484:
Removed the WIP tag. Any final corrections? All tests pass: standalone, embedded, distributed and SA.

ddebrunner:
Are you still seeing the errors in part 3 of PR #1575, because I don't think they are seen on master, so is this PR breaking them somehow?

wmarshall484:
@ddebrunner, no, all tests pass now for me after syncing with master.

wmarshall484:
All tests, Python and Java, pass testing on 4.2.

PR #1561, opened 2018-03-19
Add dependencies in hash adder key function

wmarshall484:
Currently, if the hashadder uses a third party library, the dependency will not be included automatically.


ddebrunner:
@wmarshall484 Can you enter a bug for this so it can be tracked.

ddebrunner:
-1 - as I don't see what the issue is.

The hash adder is added using `addFunctionalOperator()` which adds a dependency on its logic, the `hasher` in this case.

`hasher` is an instance of `KeyFunctionhasher` which is a `WrapperFunction` and it correctly wraps the user-supplied key function.

Issue #1543, opened 2018-03-08
Investigate optimizing conversion of SPL tuple to/from dictionary.


ddebrunner:
- [ ] Look at returning the `dict` to the C++ operators and convert it to a tuple using Python C-API.
    - [x] - `@spl.map` `@spl.source` - #1562 
    - [x] - `@spl.primitive` - #1583
    - [ ] - Functional `map`, `flat_map`, `source`
    - [ ] - Functional `aggregate`
- [ ] Look at creating a default `dict` and use that as the base when converting SPL tuple to Python `dict`. May avoid resizing of the dict as it will already have the correct number of keys.

ddebrunner:
#1562 added handling returned `dict` in the cgt code and improved performance for both `tuple` and `dict` as there is no longer one additional Python call layer. 

The change for #1562 still needs to be appled to `@spl.primitive` and functional operators.

markheger:
`flat_map` is still an open issue and covered in issue: https://github.com/IBMStreams/streamsx.topology/issues/2486

PR #1562, opened 2018-03-19
Convert dict tuples directly to SPL tuples for spl.map and spl.source.

ddebrunner:
See #1543 

For `@spl.map` and `@spl.source` the decorated callable is now called directly and all conversion of the Python object representing the tuple is handled in CCGT/C++ code.

Previously the CGT/C++ code only supported `tuple` or `list[tuple]` so any `dict` object was first converted to a `tuple` using a wrapper function.

Now a `fromPythonDictToSPLTupleN` method is created in the operator and used when the returned value is a `dict` or the value in the returned `list` is a `dict`.

A minor optimization is handled for `dict` to try avoid the cost of too many dict lookups, basically once N items have been taken from the `dict` where N is the size of the `dict` then no more lookups are performed as we have taken all the values.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/codegen/py_pyTupleTosplTuple.cgt
+++ b/com.ibm.streamsx.topology/opt/python/codegen/py_pyTupleTosplTuple.cgt
@@ -12,6 +12,7 @@
   }
 %>
  
+// Python tuple to SPL tuple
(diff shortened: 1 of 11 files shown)


ddebrunner:
`@spl.primitive` creates the conversion from `dict` method but does not yet use them, this will be added in a follow-on PR.

Topology operators do not yet support direct conversion from `dict`, they still use the wrapper approach. This will be addressed in a follow-on PR.

ddebrunner:
Passed all the Python tests.

Issue #1460, opened 2018-02-06
Support broadcast as a routing option for parallel regions.


wmarshall484:
A parallel region should support the option of having all incident tuples broadcast to every channel.

At the functional API level, this could look like:
```Java
// In Java
TStream<Integer> stream = stream.parallel(() -> 5, Routing.BROADCAST)
```

or
```python
#python
s = s.parallel(width = 5, routing = topology.Routing.BROADCAST)
```

Ref: https://www.ibm.com/support/knowledgecenter/en/SSCRJU_4.2.1/com.ibm.streams.ref.doc/doc/udpparallelannotation.html

- [x] #1542 Java api & implementation for `TStream` and SPL code generation
- [x] #1563 Support for Java `SPLStream`
- [x] #1596 Support for Python.

ddebrunner:
Just to note that this is a new feature in Streams 4.2

ddebrunner:
Also typically to be useful depends on the ability to have multiple separate inputs to a parallel region.

wmarshall484:
Notes:
* Required to support parallel streams in the Beam runner.
* Adjacent parallel regions with broadcast enabled will generate tuples exponentially since every channel broadcasts to every downstream channel.

ddebrunner:
> Adjacent parallel regions with broadcast enabled will generate tuples exponentially

**May** generate tuples exponentially, we have no idea what the first region is doing with the tuples, so the number output is not guaranteed to be order(total number in across all channels).

Not sure what should be documented, apart from the simple statement tuples are broadcast to all channels.

PR #1563, opened 2018-03-19
Add SPLStream broadcast routing support

ddebrunner:
Part of #1460


ddebrunner:
tested with streaming analytics, distributed & standalone.

PR #1575, opened 2018-03-26
Test fixes.

wmarshall484:
This PR contains 3 fixes to our unit tests:

1) We were incorrectly checking to see whether the `username` and `password` were set in a test class.

2) Two tests in `publishSubscibeTopicNames` were throwing IllegalStateExceptions instead of NullPointerExceptions.

3) The PublishSubscribeTopicNames tests were failing because the python apps they use were attempting to publish streams that were non-placeable.

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/tester.py
@@ -187,7 +187,10 @@ class Tester(object):
             raise unittest.SkipTest("Skipped due to STREAMS_DOMAIN_ID environment variable not set")
 
         test.test_ctxtype = stc.ContextTypes.DISTRIBUTED
-        test.test_config = {}
+        if hasattr(test, "username") and hasattr(test, "password"):
+            test.test_config = {'username' : test.username, 'password' : test.password}
+        else:
+            test.test_config = {}
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


ddebrunner:
I think:

1) OK.
2) OK but the reason this wasn't caught is that the class name is incorrect, it should be `PublishSubscibeTopicNamesTest` to ensure it actually runs as a unit test.
3) I think is wrong, applications should not be forced to add a dummy filter after autonomous.

ddebrunner:
@wmarshall484 Which test is failing for issue 3? I don't think it is `PublishSubscribeTopicNames`

ddebrunner:
@wmarshall484 I think 3 was fixed a couple of months ago, I'm not seeing a failure on master.

wmarshall484:
Sorry, number 3 should be the Java tests which include Python publish/subscribe apps, for example:
https://github.com/IBMStreams/streamsx.topology/blob/master/test/java/src/com/ibm/streamsx/topology/test/python/PublishSubscribeJsonPythonTest.java#L53

These tests were failing due to the Python apps calling `publish` on a non-placeable stream.

wmarshall484:
For #3, should we just make virtual operators placeable since they're going to be removed anyway?

ddebrunner:
#3 has already been fixed, so no changes needed. Besides items like Union are not placeable.

ddebrunner:
Please open independent PRs with any changes needed for #1, #2, and an issue created for item #1

PR #1588, opened 2018-04-06
Remove type information from composite invocation output port.

wmarshall484:


ddebrunner:
@wmarshall484 Is this still the approach intended, or does removing that code completely work?

Since this was not invoked for Python, it seems that maybe it's not needed?

wmarshall484:
Both Python and Java parallel tests pass.

Issue #753, opened 2017-02-14
Optimise cats cradle.


ddebrunner:
Adjacent parallel regions where the second region is hashed are not optimized due to the insertion of the Hash adder. 

This was done but was intentionally removed to allow cleanup related to optimizing topologies for 4.2 by using job config overlays to force channel isolation, rather than insertion of isolation markers (which didn't force isolation on 4.2 due to a change in parallel region fusing in 4.2).

#752 should be fixed first.

ddebrunner:
Deferring from the 1.5/1.6 release.

mrshenli:
@ddebrunner Does the conflict causing the removal of [`relocateHashAdder`](https://github.com/ddebrunner/streamsx.topology/blob/62e053adb7218def57fa74e20da82ea38f6c881a/java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java#L141) still exist? What is the conflict?

mrshenli:
Hi @ddebrunner 

I want to make sure that I understand the code correctly. I have the following question and statements.

Question:
1. Why downstream `HashAdder`s of the same `endParallel()` need to use the same routing type? [[old code]](https://github.com/ddebrunner/streamsx.topology/blob/62e053adb7218def57fa74e20da82ea38f6c881a/java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java#L172)

Are the following statements true?

2. In the current implementation, no `HashAdder` is inserted in front of `endParallel()`, so there is no need to check whether there is a `HashAdder` located immediately before `endParallel()`. [[old code]](https://github.com/ddebrunner/streamsx.topology/blob/62e053adb7218def57fa74e20da82ea38f6c881a/java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java#L213)

3. In the current implementation, it is always `HashAdder` -> `Parallel` -> `HashRemover`, instead of `HashAdder` -> `Parallel` -> `Isolate` -> `HashRemover` as checked in the [[old code]](https://github.com/ddebrunner/streamsx.topology/blob/62e053adb7218def57fa74e20da82ea38f6c881a/java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java#L197).

[...]

PR #1589, opened 2018-04-06
Remove relocate hash adder dead code

ddebrunner:
Remove old code for relocating hash adders that is probably not up to date with new composite generation and did not support Python either.

Optimization of adjacent parallel regions is being tracked by #753


ddebrunner:
Ran Java `ParallelTest` standalone and distributed.

PR #1591, opened 2018-04-09
Add support for set_parallel in Python.

wmarshall484:


ddebrunner:
Shouldn't `set_parallel` return `self`?

wmarshall484:
Added version tag, now returns correct value, and added docstring. Also updated that .pyi file.

PR #1592, opened 2018-04-10
Move HashAdder/Remover to SPI

ddebrunner:
A step towards allowing a toolkit using the SPI to invoke parallelism.

Java `ParallelTest` run in standalone/distributed.


ddebrunner:
Also moves the Java `HashRemover` and `PassThrough` operators to the SPI and adds an `spi.properties` to allow the namespace of the functional Java operators to be set. This then allows an SPI to provide its own versions (extending the ones from this toolkit) that will share a class loader with its own functional operators.

PR #1600, opened 2018-04-13
Run SPL Python primitive operator extraction when building remotely

### LINKED ISSUE #1416: Include spl-python-extract in remote builds.
When an SPL build(*) (and its toolkits) are submitted for execution call the SPL Python operator extraction script for each toolkit when Python operators might exist (`opt/python/streams` exists).

(*) Could also be a Python app that included SPL Python operators!

ddebrunner:
Performs two actions on a remote build if an SPL toolkit contains SPL primitive operators:

- Install any required packages into the build output directory from `opt/python/streams/requirements`
- Run the `spl-python-extract.py` script ensuring any installed packages are available.

This then allows CI systems or any remote build use (e.g. streamsx-runner) to build and run tests/jobs including a toolkit with SPL Python primitive operators without requiring a local Streams install to perform operator extraction before submission.

This fixes #1416


ddebrunner:
Passed the new test and am running all the streaming analytics tests under test/python/topology and will do a spot check on Java tests using a remote build

PR #1605, opened 2018-04-16
Allow a topology to use pip to install dependent packages.

### LINKED ISSUE #1604: Allow packages to be installed by pip by the build server.
Allow a topology to depend on packages that will be installed by pip during a build using Streaming Analytics build service.

This ensures that the package and any dependencies are installed correctly including the correct version of any shared libraries that are platform dependent. The current dependency discovery mechanism can fail if a dependency is missed or the local system where the graph is declared is not the same os version and/or architecture as Streaming Analytics service.

For example.

```
topo = Topology()
topo.add_pip_package('pint')
topo.add_pip_package('astral==0.8.1')
```

ddebrunner:
Fixes #1604 

Dependent on #1600 so no tests run yet. It uses the mechanism added in #1600 by creating a `opt/python/streams/requirements.txt` from the requirement specifiers added using `add_pip_package`.


ddebrunner:
`test2_pkg.py` tests pass now #1600 was merged.

PR #1612, opened 2018-04-19
Fetch app logs in DISTRIBUTED context when possible.

wmarshall484:


wmarshall484:
Needs to be tested, and the warning message could be improved for when `always_fetch_logs` is set on a DISTRIBUTED submission that doesn't support fetching logs.

ddebrunner:
The docstrings for the retrieve methods should document that `None` can be returned.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -561,8 +561,10 @@ def retrieve_log_trace(self, filename=None, dir=None):
 
         if not filename:
             filename = _file_name('job', self.id, '.tar.gz')
- 
-        return self.rest_client._retrieve_file(self.applicationLogTrace, filename, dir, 'application/x-compressed')
+        if hasattr(self, "applicationLogTrace") and self.applicationLogTrace is not None:
```
ddebrunner: Ideally the creating a default `filename` would occur only if the method is going to retrieve logs.

ddebrunner:
Multiple methods use `_retrieve_file`, they should all have similar checks.

PR #1615, opened 2018-04-23
Add examples for fetching REST objects using regular expressions.

wmarshall484:


ddebrunner:
Examples need to be explained a bit more, as to what the example is trying to show. Ie. the reader should look at the code understanding what it is doing, not have to figure out what it is doing from the code.

Don't use Condition and the metic name example.

I think for views, the main concern has been how to use the View object to get data, not just how to get a view object.

PR #1639, opened 2018-05-01
Ensure conflicting parallel starts fail.

wmarshall484:


wmarshall484:
STP have been taken into account.

PR #1650, opened 2018-05-07
Add pyStateful parameter to topology python operators

EricMichaelJohnson:


EricMichaelJohnson:
I accidentally added a commit that should not be part of this pull request.  I will close it and create a new one with the right contents.

PR #1653, opened 2018-05-08
Added the pyStateful parameter to python topology operators.

EricMichaelJohnson:
I added a parameter, pyStateful, to the python topology operators.  It indicates whether on operator has state that needs to be saved when checkpointing.  All source operators are stateful, and other operators are stateful if their callable is not a function.

### DIFF
--- a/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/Aggregate/Aggregate.xml
+++ b/com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/Aggregate/Aggregate.xml
@@ -100,9 +100,9 @@
       <parameter>
         <name>pyStateful</name>
         <description>Whether the operator has state to be saved in checkpointing.</description>
-        <optional>true</optional>
+        <optional>false</optional>
         <rewriteAllowed>true</rewriteAllowed>
-        <expressionMode>AttributeFree</expressionMode>
+        <expressionMode>Constant</expressionMode>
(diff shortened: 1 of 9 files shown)


wmarshall484:
> operators are stateful if their callable is a function.

Per the `determine_statefullness` method, I think you mean operators are stateful if their callable is not a function.
```
 def _determine_statefulness(self, _callable):
        stateful = not inspect.isroutine(_callable)
        return stateful
```

EricMichaelJohnson:
@wmarshall484, you are correct.  I edited the comment.

ddebrunner:
Why are the parameters optional, when would they not be set?

Why is the mode `AtttributeFree` and not `Constant`?

ddebrunner:
If all source operators are stateful, why does `Source` have an `pyStateful` parameter?

EricMichaelJohnson:
`Source` has a `pyStateful` parameter for two reasons: it prevents source having to be treated as a special case in the operator generation, and it will be easier to change if we find a case in which `Source` should not be stateful.

EricMichaelJohnson:
I tested the simple lambda expression in the filter operator in the filter_echo sample, and it was correctly determined not to be stateful.

EricMichaelJohnson:
I updated the operators to have Constant mode and to be optional.

ddebrunner:
I will merge this once some other fixes have gone into master that are needed for 1.9/1.10.
(this is post 1.10 work).

PR #1670, opened 2018-05-17
Fix #1668 - When used from strreamsx package find correct toolkit

### LINKED ISSUE #1668: pip version of 1.9.3 beta finds wrong toolkit directory in spl-python-extract script
I am using Python 3.6.5.

I notice in `spl-python-extract`, the toolkit directory is identified by the location of the source file, and then 5 calls to `dirname(dir)`. On the version installed by `pip3.6 install streamsx==1.9.3b`, the result is incorrect, resulting in none of the localization files being copied correctly.

It returns the toolkit directory as being `"/usr"`, so it looks for the TopologySplpyResource definitions in `"/usr/impl/nl"`.

I found that I can workaround this by changing `extract.py` to have a hard-coded return value for that function:

    return "/usr/lib/python3.6/site-packages/streamsx/.toolkit/com.ibm.streamsx.topology"

I'm not sure of the best solution, but an option may be to check for the presence of a `.toolkit` directory when you traverse up the tree with `dirname`.

ddebrunner:


ddebrunner:
Tested with `tests/python/spl/tests` unittests.

PR #1692, opened 2018-06-01
Checkpoint

EricMichaelJohnson:


ddebrunner:

ddebrunner:
run tests

EricMichaelJohnson:
Created by mistake.  I was trying to pull in the other direction.

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
```diff
@@ -114,6 +117,58 @@ class SplpyFuncOp : public SplpyOp {
...
+        // a state handler instance.  Otherwise, register a do-nothing
+        // state handler.
+        SPL::boolean stateful = static_cast<SPL::boolean>(op()->getParameterValues("pyStateful")[0]->getValue());
+        assert(!stateHandler);
+        // Ensure that callable() can be pickled before using it in a state
+        // handler.
+        PyObject * pickledCallable = NULL;
+        if (stateful) {
+          SplpyGIL lock;
+          PyObject * dumps = SplpyGeneral::loadFunction("dill", "dumps");
```
ddebrunner: Is it possible to use the existing generic function call function?

https://github.com/IBMStreams/streamsx.topology/blob/master/com.ibm.streamsx.topology/opt/python/include/splpy_general.h#L354

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
```diff
@@ -114,6 +117,58 @@ class SplpyFuncOp : public SplpyOp {
...
+              PyObject * type = NULL;
+              PyObject * value = NULL;
+              PyObject * traceback = NULL;
+
+              PyErr_Fetch(&type, &value, &traceback);
+              if (value) {
+                SPL::rstring text;
+                // note pyRStringFromPyObject returns zero on success
+                if (!pyRStringFromPyObject(text, value)) {
+                  SPLAPPTRC(L_WARN, "Checkpointing is disabled for the " << op()->getContext().getName() << " operator because of python error " << text, "python");
```
ddebrunner: Seems like this is an error, not just a warning.

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
```diff
@@ -131,6 +186,92 @@ class SplpyFuncOp : public SplpyOp {
...
+           throw SplpyGeneral::pythonException("dill.loads");
+         }
+         Py_DECREF(op->callable());
+         op->setCallable(initialCallable);
+       }
+
+      private:
+        // Call a python callable with a single argument
+        // Caller must hold GILState.
+        PyObject * call(PyObject * callable, PyObject * arg) {
```
ddebrunner: We already have utility methods to call objcets:

https://github.com/IBMStreams/streamsx.topology/blob/master/com.ibm.streamsx.topology/opt/python/include/splpy_general.h#L162

and at line 181

com.ibm.streamsx.topology/opt/python/include/splpy_general.h
```diff
@@ -187,7 +187,36 @@ class SplpyGeneral {
...
     }
 
-    
+    /**
+     * Utility method to write a python exception to the application trace.
+     * The type and value are written, but not the traceback.
+     * If no python exception occurred, this does nothing.
+     * The caller must hold the GILState.
+     */
+    static void tracePythonError() {
```
ddebrunner: We already trace Python errors, so why is a new function needed?

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.pyi
```diff
@@ -31,6 +31,7 @@ class Topology(object):
     def add_file_dependency(self, path: str, location: str) -> str: ...
     def add_pip_package(self, requirement: str) -> None: ...
     def create_submission_parameter(self, name:str, default=None, type_=None): ...
+    def checkpoint_period(self, period: int, uints: TimeUnit) -> None: ...
```
ddebrunner: Signature is wrong, needs to be a union of float & datetime.timedelta

samples/java/functional/src/simple/Echo.java
```diff
@@ -53,6 +57,7 @@
     public static void main(String[] args) throws Exception {
 
         Topology topology = new Topology("Echo");
+	topology.checkpointPeriod(1, TimeUnit.MINUTES);
```
ddebrunner: Seems to be debugging code left in??

samples/python/topology/simple/echo.py
```diff
@@ -28,6 +29,8 @@ def main():
     """
     
     topo = Topology("echo")
+    topo.checkpoint_period = timedelta(seconds=1)
```
ddebrunner: Add a new sample if a checkpoint sample is needed, echo should be very simple.

samples/python/topology/spl/files.py
```diff
@@ -38,7 +38,7 @@ def main():
     lines.print()
     
     # Now execute the topology by submitting to a standalone context.
-    streamsx.topology.context.submit("STANDALONE", topo)
+    streamsx.topology.context.submit("TOOLKIT", topo)
```
ddebrunner: Leave as standalone.

samples/python/topology/simple/filter_echo.py
```diff
@@ -26,6 +27,7 @@ def main():
     """
     
     topo = Topology("filter_echo")
+#    topo.checkpoint_period(1, TimeUnit.MINUTES)
```
ddebrunner: Don;t change this existing examples .

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
EricMichaelJohnson: Really?  If someone enables checkpointing in a topology with an operator that cannot be saved, we don't allow it to run at all?

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
EricMichaelJohnson: The way that the existing callFunction method works is that on failure, it writes the python error information (including a traceback into dill code that is not likely to be useful to us or our users) into standard error, consumes the python error, and then throws an exception.  Using our application tracing, messages are written with timestamps, with the name of the operator in which they occurred, and the location in our source where they occurred.  The messages will also be shown in the correct sequence of the other trace messages.  I think without that information it will be difficult to diagnose the causes of python exceptions that may happen.

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
EricMichaelJohnson: The callFunction method in splpy_general log python errors to standard error, with no information about when they happened, in what operator, and from where in our code.  Diagnosing consistent region issues is very tricky, and I think it is important to log any problems in the application trace to preserve information about what was happening when an error occurred.

com.ibm.streamsx.topology/opt/python/include/splpy_general.h
EricMichaelJohnson: Python errors are written to standard error, but not the application trace.  This will not preserve important information for diagnosing problems, such as the time the error occurred and the operator it occurred in.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.pyi
EricMichaelJohnson: I forgot to update that.  What do you mean by a union of float and datetime.timedelta?  Is there some support for c-like union data types in python?

samples/python/topology/spl/files.py
EricMichaelJohnson: Sorry, I created the pull request by mistake before finishing my review.  I will change this.

samples/python/topology/simple/filter_echo.py
EricMichaelJohnson: Sorry, unintended change.  I will change it back.

EricMichaelJohnson:
This was accidentally created prematurely, but since Dan has already reviewed it we may as well go forward with this one.

com.ibm.streamsx.topology/opt/python/include/splpy_funcop.h
ddebrunner: Well, it's not fulfilling the required functionality. Seems safer to fail rather than continue on in silence not checkpointing.

EricMichaelJohnson: I updated to log an error and throw an exception, shutting down the operator.

ddebrunner:
run tests

ddebrunner:
Python tests saw several failures such as this, even though I believe these tests are not checkpointing.

This was in `test2_udp.py`
```
05 Jun 2018 12:46:41.951 [15061] ERROR #splapptrc,J[0],P[0],StreamContents0,spl_operator M[PEImpl.cpp:handleOperatorFailure:646]  - CDISR5030E: An exception occurred during the execution of the StreamContents0 operator. The exception is: Checkpointing is not available for the StreamContents0 operator because of python error "dictionary changed size during iteration"
```

EricMichaelJohnson:
I fixed the failures in test2_udp.py.

ddebrunner:
jenkins run tests

Issue #1698, opened 2018-06-05
Align exception handling with existing code.


ddebrunner:
Follow on to #1692 @EricMichaelJohnson

We should be handling exceptions consistently including passing the exception to `__exit__` if it exists and suppressing it if needed.

If improvements are needed in the existing mechanism then they should be added there.

EricMichaelJohnson:
This issue can be closed.

ddebrunner:
There still seems to be multiple exception handling calls, each doing something different.

markheger:
no hint what is missing, close issue

PR #1708, opened 2018-06-06
Checkpoint 1698

EricMichaelJohnson:
Enable the user's class to suppress the exception thrown when the user's callable object cannot be dilled.  When this happens, processing continues with no checkpoint support for the operator using the callable that cannot be dilled.


ddebrunner:

EricMichaelJohnson:
I did some ad-hoc testing to ensure that this works with exceptions suppressed and not suppressed.  I can add automated tests.

ddebrunner:
jenkins run tests

ddebrunner:
Why is the documentation being removed from `checkpoint_period`?

EricMichaelJohnson:
fixes #1698

PR #1710, opened 2018-06-06
Some safety fixes to SplpyExceptionInfo

EricMichaelJohnson:
The default constructor for SplpyExceptionInfo leaves et_ and location_ uninitialized, potentially leading to undefined behaviour if the uninitialized values are used.  It appears that the intent of this class is that the default constructor is not to be used to construct an instance of the class, but instead the static methods exception and dataConversion, so I made the default constructor private.  asTuple had a potential buffer overrun if pyTraceback_ is set but not pyValue_, which I think is possible.


EricMichaelJohnson:
jenkins run tests

ddebrunner:

PR #1723, opened 2018-06-13
Checkpoint exception tests

EricMichaelJohnson:
Adds tests for successful runs with checkpointing enabled in a python topology operator.  This includes a case where the callable can be dilled, and cases where it cannot be dilled, with and without suppressing the exception thrown when attempting to dill the callable.


ddebrunner:

PR #1733, opened 2018-06-21
Fix samples property for test with changed toolkit version

joergboe:
If the test is called with a changed toolkit release (e.g. from Streams installation) the samples have a different location. Thus it must be possible to setup the samples-jar independent from toolkit release.
This should also be pulled into v1_10


ddebrunner:

PR #1738, opened 2018-06-25
Relocate HashAdder (Step 1)

mrshenli:
Relocate `HashAdder` for a special case `$Unparallel$ -> HashAdder -> $Parallel$ -> HashRemover` where `HashAdder` is `$Unparallel$`'s only child, and `$Unparallel$` is `HashAdder`'s only parent.

Related to #753


java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +71,45 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
     }
+
+    @SuppressWarnings("serial")
+    private void relocateHashAdders(){
+        final Set<JsonObject> hashAdders = new HashSet<>();
+        // Firstly, find each hashAdder
+        GraphUtilities.visitOnce(GraphUtilities.findStarts(graph),
+                new HashSet<BVirtualMarker>(),
+                graph,
+                new Consumer<JsonObject>(){
```
ddebrunner: We usually use a lambda expression rather than anonymous class.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +71,45 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+                    public void accept(JsonObject op) {
+                        if(jstring(op, "kind").equals("com.ibm.streamsx.topology.functional.java::HashAdder")){
```
ddebrunner: This is insufficient, and one of the reasons the old code was removed. What is actually needed is a property of the operator invocation to mark it as a logical hash adder, and not determine it from its kind.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +71,45 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+    private void relocateHashAdder(JsonObject hashAdder){
+        // $Unparallel$ -> hashAdder -> $Parallel$ -> hashRemover
+        Set<JsonObject> parents = GraphUtilities.getUpstream(hashAdder, graph);
+        
+        // hashAdder has multiple parents
+        if (parents.size() != 1) return;
+
+        JsonObject parent = parents.iterator().next();
+
+        // hashAdder's parent has multiple children
```
ddebrunner: How could a hash adder have multiple downstream children?

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +71,45 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+        if (GraphUtilities.getDownstream(parent, graph).size() != 1)
+            return;
+
+        // Only optimize the case where $Unparallel has only one child, and
+        // $parallel has only one parent
+        if (jstring(parent, "kind").equals(BVirtualMarker.END_PARALLEL.kind())) {
```
ddebrunner: There are utility methods in graph operators to get the kind of an operator invocation.

test/java/src/com/ibm/streamsx/topology/test/api/ParallelTest.java
```diff
@@ -78,8 +78,7 @@ public void fanoutEndParallelException() throws Exception {
 	assertTrue(expectedCount.valid());
     }
 
-    @Test
-    public void testAdjacentParallel() throws Exception {
+    public void testAdjacentParallel(Routing routing1, Routing routing2) throws Exception {
```
ddebrunner: private method - only used in this test.

.gitignore
```diff
@@ -43,3 +43,9 @@ samples/java/functional/classes/
 samples/java/functional/functionalsamples.jar
 /.project
 /.pydevproject
+
```
ddebrunner: Please don't include unrelated changes in a pr. Submit them as their own pr.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
mrshenli: The **parent** of a `HashAdder` can have multiple children.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
mrshenli: Should I add that property to `OpPerperties` and `BOperatorInvocation#_json()` similar to what is done for `KIND`? Any suggestion for the property name? What about `HASHADDER_OP`?

ddebrunner: Boolean JSON property "hashAdder" added in OpProperties.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
ddebrunner: I think I'm confused by the description of this PR which has:

> where HashAdder is $Unparallel's only child **and $parallel's only parent.**

So I thought the check was that the $parallel only has a single upstream operator, which is the hashadder, which will always be the case.

ddebrunner: Could you add comments as to why you are putting those restrictions in place, it's fine if it's because only the simple case is being tackled at the moment, but the code is much clearer if a reason is given.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +71,45 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+        if (parents.size() != 1) return;
+
+        JsonObject parent = parents.iterator().next();
+
+        // hashAdder's parent has multiple children
+        if (GraphUtilities.getDownstream(parent, graph).size() != 1)
+            return;
+
+        // Only optimize the case where $Unparallel has only one child, and
+        // $parallel has only one parent
```
ddebrunner: See, again "$parallel has only one parent" but that's not what the code is checking for, so is the code correct and the comment wrong, or vice-versa?

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
mrshenli: Sorry, I should have updated the PR description.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
mrshenli: Do you mean this [`OpProperties`](https://github.com/IBMStreams/streamsx.topology/blob/master/java/src/com/ibm/streamsx/topology/generator/operator/OpProperties.java)? I didn't see a `"hashAdder"` property.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
ddebrunner: That's probably the comment I was looking for (assuming it is corrected), but it should be before the checks so that a reader understands the checks as they read down.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +71,45 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+        if (jstring(parent, "kind").equals(BVirtualMarker.END_PARALLEL.kind())) {
+            GraphUtilities.removeOperator(hashAdder, graph);
```
ddebrunner: Does `removeOperator` followed by `addOperator` do the right thing. I think previously `removeOperator` was only used to remove operators that were to be discarded, not added back into the graph. I can't see if the connections for the resultant operator are correct, i.e. does the operator end up with more connections that it should have?

mrshenli: this is not addressed yet

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
mrshenli: this is not addressed yet

mrshenli:
@ddebrunner 

> Boolean JSON property "hashAdder" added in OpProperties.

I am confused about the above. Do you mean the "hashAdder" property has already been added? I did not find it in [`OpProperties`](https://github.com/IBMStreams/streamsx.topology/blob/master/java/src/com/ibm/streamsx/topology/generator/operator/OpProperties.java). Or do you mean I should add a "hashAdder" property to `BOperatorInvocation#_json()`?

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +70,46 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+        // parent and the HashAdder is $Unparallel$'s only child.
+        // $Unparallel$ -> hashAdder -> $Parallel$ -> hashRemover
+        Set<JsonObject> parents = GraphUtilities.getUpstream(hashAdder, graph);
+        
+        // check if hashAdder has only one parent
+        if (parents.size() != 1) return;
+
+        JsonObject parent = parents.iterator().next();
+
+        // check if $Unparallel$ has only one child
```
ddebrunner: FYI - somewhat confusing here because at this point (line 104) it's not guaranteed to be  `$Unparallel$`, since you only check that at line 108. It would actually make more sense to check first it's an `$Unparallel$` before seeing how many downstream operators it has.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +70,46 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
         for (JsonObject composite : composites)
             pePlacementPreprocess.compositeColocateIdUse(composite);
     }
+
+    @SuppressWarnings("serial")
+    private void relocateHashAdders(){
+        final Set<JsonObject> hashAdders = new HashSet<>();
+        // Firstly, find each hashAdder
+        GraphUtilities.visitOnce(GraphUtilities.findStarts(graph),
```
ddebrunner: Is there a reason to collect the list of hash adders first, rather than just relocate them as you discover them?

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +70,46 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
         for (JsonObject composite : composites)
             pePlacementPreprocess.compositeColocateIdUse(composite);
     }
+
+    @SuppressWarnings("serial")
```
ddebrunner: Is the suppress warnings needed?

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
```diff
@@ -68,4 +70,46 @@ public void compositeColocateIdUsage(List<JsonObject> composites) {
...
+
+        JsonObject parent = parents.iterator().next();
+
+        // check if $Unparallel$ has only one child
+        if (GraphUtilities.getDownstream(parent, graph).size() != 1)
+            return;
+
+        if (GraphUtilities.isKind(parent, BVirtualMarker.END_PARALLEL.kind())) {
+            JsonObject hashAdderCopy = GraphUtilities.copyOperatorNewName(
+                    hashAdder, jstring(hashAdder, "name") + "_0");
```
ddebrunner: No need to generate a new name, since you are only moving the operator.

java/src/com/ibm/streamsx/topology/generator/spl/Preprocessor.java
ddebrunner: probably is, due to to not being able to handle graph changes during the iteration.

PR #1740, opened 2018-06-25
Checkpoint 1711

EricMichaelJohnson:
Allow checkpoint_period to be either a datetime.timedelta, or any value that can be cast to a float.  Also enforce a minimum checkpoint_period of 0.001 seconds, which is the lowest value the runtime supports.


EricMichaelJohnson:
Meant to be for feature/consistent, not master.

Issue #1711, opened 2018-06-07
checkpoint_period should support passing an int


ddebrunner:
Or maybe if the value is not a datetime.timedelta, try converting to float and use that value if it succeeds (more Pythonic).

EricMichaelJohnson:
This issue can be closed.

PR #1741, opened 2018-06-25
Checkpoint 1711

EricMichaelJohnson:
Allow checkpoint_period to be either a datetime.timedelta, or any value that can be converted to a float.  Also enforce a minimum checkpoint_period of 0.001 seconds, which is the lowest value supported by
the runtime.  This  is related to issue #1711


EricMichaelJohnson:
fixes #1711

PR #1745, opened 2018-06-27
Add initial CI Python testing

ddebrunner:


ddebrunner:
Correctly runs the test, test that was failing is fixed by #1764

PR #1762, opened 2018-07-05
fix Unparallel's input port type after relocating HashAdder

mrshenli:


java/src/com/ibm/streamsx/topology/generator/spl/GraphUtilities.java
```diff
@@ -533,7 +533,17 @@ static void setOutputPortType(JsonObject op, int index, String schema) {
...
     }
-    
+
+    /**
+     * set the input port schema type to the given value
+     */
+    static void setInputPortType(JsonObject op, int index, String schema) {
+        JsonArray inputs = op.get("inputs").getAsJsonArray();
+        JsonObject input = inputs.get(index).getAsJsonObject();
+        input.remove("type");
```
ddebrunner: There's no need to have a remove of the property is there?

mrshenli: no, it is not necessary. I was hesitated to use the `addProperty` API here because it is actually doing an update/set. Let me change the code to use `addProperty`.

PR #1765, opened 2018-07-12
Support for pickled decorated operators.

ddebrunner:


ddebrunner:
Went through one complete test cycle.

PR #1776, opened 2018-07-20
[WIP] Port unique ids

ddebrunner:


ddebrunner:
Going to take a different approach, using in the name as the unique identifier to avoid too many changes.

PR #1779, opened 2018-07-23
Add simplified layout kinds for Java functional ops

ddebrunner:


ddebrunner:
Python tests (not affected by this change) failed due to resource issues on test machine.

PR #1786, opened 2018-08-01
Fix Python includes for primitive operators

ddebrunner:
Previously the includes were working by chance as `python-config -includes` returned duplicated paths like:

`-I/python3/include -I/python/include`

which was then handled by the SPL compile system as two tokens, the first an **relative** include path and the second just a flag to the compiler, resulting in C++ flags like (simplified):

`-I/topology_tk_path/.../Map/-I/python3/include -I/python/include`

The script that produces the paths is expected to have one path per-line of output.


ddebrunner:
Test failures are related to resource issues on the test machine.

PR #1832, opened 2018-08-23
Try marking build as unstable on Python failure.

ddebrunner:


ddebrunner:
Correctly marks build as unstable.

PR #1835, opened 2018-08-24
Disable xunit from nose as it does not work with processes

ddebrunner:


ddebrunner:
Test failures are unrelated to this change - on-going issue with test_exceptions.

PR #1852, opened 2018-08-30
Ensure stack trace is available

ddebrunner:
The stack trace must be available to help with debugging.

Without this I was getting only the info 'Internal Error'. With it I could see the full stack trace and instantly understand where the issue was.


EricMichaelJohnson:
Logging the exception type and value to the application trace along with the traceback to standard error should enable one to correlate the traceback to the message in the application trace, and recover the information about when it happened, what operator, what thread, etc.

PR #1874, opened 2018-09-08
Fix no-tuples count conditions.

ddebrunner:


ddebrunner:
Unrelated test failure.

PR #1921, opened 2018-10-09
[WIP] Aggregate consistent region protocol.

ddebrunner:
1) Drain window during consistent region processing at drain time.
2) Improve test to have state maintained in aggregate logic.


com.ibm.streamsx.topology/com.ibm.streamsx.topology.functional.python/Aggregate/Aggregate_cpp.cgt
```diff
@@ -225,22 +225,30 @@ void MY_OPERATOR::onWindowTriggerEvent(Window<PyObject *> & window, Window<PyObj
 // End Window Event Handler Overrides
 // ##################################
 
+#if SPLPY_OP_STATEFUL == 1
```
EricMichaelJohnson: Won't SPL_OP_STATEFUL always be 1, so the CPP intrusion is not needed?

ddebrunner: It will be 0 if the operator is not checkpointing and not in a consistent region.

PR #1941, opened 2018-10-22
Locking using existing RAII constructs.

ddebrunner:
Locking of Python operators using existing `AutoMutex` or `AutoPortMutex` RAII classes.

Adds in locking of stateful operators when not using a state handler using `AutoPortMutex`. This provides cleaner code for Python developers who need not worry about locking in most cases.

Replaces `OptionalAutoLock`, `DelegatingStateHandler` for these reasons:

1. Single lock api is not sufficient for all cases, `Aggregate` locking is tricky
1. Needed to support locking when not using statehandler with `AutoPortMutex`
1. Use of existing Streams paradigms `AutoMutex`,  `AutoPortMutex` & `StateHandler` directly implemented by operator makes it in line with Streams documentation, other operators (e.g. SPL toolkit) and is clearer for other developers looking at the code.
1. Reduces amount of code to be compiled by gcc (trying to reduce compilation time) by not having classes that are not used but still being compiled. 
1. State handling of the callable associated with `Aggregate` is done through `Window.registerSerializer` callbacks which then handles locking correctly when time based policies are used.

Still to be done, wanted to get an early test run:

- [x] Aggregate operator
- [x] SPL decorated Source operator
- [x] Removal of `OptionalAutoLock`,`DelegatingStateHandler`
- ~Add checkpoint/consistent region variants of mt (multi-threaded) tests.~ Tests will be added in #1967


ddebrunner:
Passed Python checkpointing/consistent region tests for topology and SPL operators with the service.

EricMichaelJohnson:
I am sure we disagree about this, but I think for dynamic code generation in modern C++ code, templates should be preferred to the old c preprocessor.

ddebrunner:
Merging to be part of the beta as it fixes some locking issues for cr.

@EricMichaelJohnson If you have specific improvements please enter an issue.

Issue #1866, opened 2018-09-05
The default ant target 'all' fails in python sample com.ibm.streamsx.topology.pysamples


joergboe:
With commit [3eff52c](https://github.com/IBMStreams/streamsx.topology/commit/3eff52c9739cdd168534f31b6b0508e3da24eaae) the the default target 'all' is broken:

> Buildfile: /home/joergboe/git/streamsx.topology/samples/python/com.ibm.streamsx.topology.pysamples/build.xml
> [echo] Build the toolkit with topology toolkit in /opt/ibm/InfoSphere_Streams/4.3.0.0/toolkits/com.ibm.streamsx.topology
> 
> spldoc:
> [exec] CDISP9209E ERROR: The /home/joergboe/git/streamsx.topology/samples/python/com.ibm.streamsx.topology.pysamples/impl/nl/en_US/TopologySplpyResource.xlf file was not found.
> [exec] CDISP0786E ERROR: The generation of the TopologySplpyResource resource bucket failed.
> [exec] CDISP9228E ERROR: The spl-make-toolkit command run by the spl-make-doc command failed. The return code is: 1.
> 
> BUILD FAILED
> /home/joergboe/git/streamsx.topology/samples/python/com.ibm.streamsx.topology.pysamples/build.xml:55: exec returned: 1

This issue is a side effect from #1706 
The problem:
This sample is a mixture of spl code and python operators. Thus is has defined an info.xml file. The extraction of the python operators with spl-python-extract.py generates the 'impl/nl' resource files and checks the info.xml of the toolkit.
To avoid manual interaction of the user, the info.xml of the sample has already the resource section impl/nl/TopologySplpyResource, thus the spl-make-toolkit fails if there was no run of spl-python-extract.py beforehand. So the original version of the build.xml had a dependency of the spldoc target to target samples.

Possible solutions:
1. Remove the info.xml from the sample source: The info.xml is generated during python operator extraction, but then the general description is empty and the genrated python operators are not in spldoc.

2. The 'spldoc' target must depend (at least) on target 'toolkit'
But in this case the release bundle contains the extracted python-operators, which is probably also not correct.

3. Introduce a new target 'clean-toolkit' which must be called after 'spldoc', when the release bundle is made.

joergboe:
Corrected with  PR #2089

PR #2089, opened 2019-04-03
Fix issue #1866

joergboe:
In the sample/python/com.ibm.streamsx.topology.pysamples/build.xm :
the spldoc-target depends now on the-toolkit which extracts the python primitive operators


ddebrunner:
@joergboe FYI - the recommended development model is forking `streamsx.topology`, not creating shortlived branches in the main repo.

https://github.com/IBMStreams/streamsx.topology/blob/master/DEVELOPMENT.md

Issue #2093, opened 2019-04-09
New Python Wrapper Functions required for Publish and Subscribe


joergboe:
For a special application we need 2 new python wrapper functions for publish and subscribe. These functions read the required parameters for the publish and subscribe operator from a special dictionary. This dictionary contains among other information:
- Topic
- Schema,
- Connection mode
- Buffer Capacity
- Buffer Full Policy

joergboe:
If we pass the dictionary as **params, it must have exactly the key value pairs that are expected from the function. But this is not the case. The special connection dictionary contains the required parameters among others (e.g. Connetion name,  Instance name...) Thus we need a function as adapter adapter from connection dictionary to publis/subscribe function.

markheger:
@joergboe further PR or can this issue be closed?

joergboe:
Due to incomplete and unclear requirements, this issue can be closed.

PR #2096, opened 2019-04-10
New Python Wrapper Functions required for Publish and Subscribe issue #2093

joergboe:
This PR is available for discussion.

See #2093

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/context.py
@@ -877,6 +877,27 @@ class ConfigParams(object):
     .. seealso:: :ref:`sas-service-def`
     """
 
+    SC_OPTIONS = 'topology.sc.options'
+    """
+    Options to be passed to IBM Streams sc command.
+
+    A topology is compiled into a Streams application
+    bundle (`sab`) using the SPL compiler ``sc``.
+
+    Additional options to be passed to ``sc``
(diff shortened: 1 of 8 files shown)


ddebrunner:
I'm reluctant to increase the surface area of the api to handle special cases. Most likely this can be done with a decorator on the `publish` and `subscribe` operators that effectively implements the last solution (by Aviendha) in this: https://stackoverflow.com/questions/26515595/how-does-one-ignore-unexpected-keyword-arguments-passed-to-a-function

If we do add additional methods then a dictionary comprehension would be a one liner that replaces most of the code from 592 to 607 and 1754 to 1759 , e.g.

```
ca = {k: connection[k] for k in ('topic', 'schema', 'connection_mode', 'buffer_capacity', 'buffer_full_policy') if k in connection}
self.subscribe(**ca)
```

ddebrunner:
I also wonder if instead of having the connection definition use 'Python', 'String' why not just always use the SPL schemas, and let clients decide how to map it into their local representation.

At least `subscribe` and `subscribe_connection` should accept the same values for schema.

joergboe:
I assume that the core module, which generated the connection dictionary is not dependent from all the toolkit modules. Thus I assume all connection values are strings.

ddebrunner:
I mean the string version of the raw SPL schema, e.g. `tuple<rstring string>` rather than `String`, but maybe the logical version is ok, just a pure SPL tool would have to do a conversion.

PR #2104, opened 2019-05-14
partitioning python topology windows by attribute

EricMichaelJohnson:
Support for partitioned windows in python topology operators.  This contains support for partitioned windows in operator with structured schemas, partitioning by a single attribute in the schema.


ddebrunner:
`def partition(self, attribute):`

How will partition by expression be supported?

Should the method take a `key` instead, which may be an attribute name (`str`) or a callable in the future?

EricMichaelJohnson:
I am already working on support for partitioning using a callable. Yes, it would probably be better to call the argument to `partition` `key` rather than `attribute`.  I will make that change in the code I am working on.

PR #2116, opened 2019-06-05
WIP python windows partitioned by python callable.

EricMichaelJohnson:


EricMichaelJohnson:
This has a pull from master, which we don't want.

PR #2119, opened 2019-06-06
Support for python windows partitioned by a python callable.

EricMichaelJohnson:


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -2109,8 +2142,16 @@ def partition(self, attribute):
         Returns:
             Window: Window that will be triggered.
         """
+
+        if 'partitioned' in self._config and self._config['partitioned']:
```
ddebrunner: Since a new window is returned is this check needed? The partitioning replaces any existing partitioning.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -2079,19 +2082,49 @@ def _evict_time(self, duration):
...
+            dilled_callable = base64.b64encode(dill.dumps(function, recurse=recurse)).decode("ascii")
+
+        self._config['partitioned'] = True
+        if dilled_callable is not None:
+            self._config['partitionByCallable'] = dilled_callable
+        self._config['partitionByName'] = name
+        self._config['partitionByModule'] = function.__module__
 
-    def partition(self, attribute):
+    def partition(self, key):
```
ddebrunner: Docstring has not been updated for this change, lines 2135, 2140, 2143.

Can you also add a `versionadded` to the docstring with version 1.13.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
EricMichaelJohnson: I think it would be unexpected for the user for multiple calls to partition to be accepted, but only one of them to be applied.  I think if someone calls partition multiple times, it is probably either an error, or the user expects all of the calls to apply.  Failing with an error message helps the user understand what will be applied in these cases.

EricMichaelJohnson:
There is one test that uses a function to partition, but I will add more.

Currently there is no support for saving the state of a stateful partition function in a consistent region, but it does not look too difficult to add that.

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
ddebrunner: Each one would be applied, each partition call returns a new window. That needs to be clear from the docstring. See `trigger`.

EricMichaelJohnson:
I have added support for checkpointing the callable function (but not the context manager), and addressed the previous review comment.

PR #2142, opened 2019-06-19
 implemented basic functionality for rm, mk, ch, and get - appconfig, still need to test

Karsten12:


ddebrunner:
Some cleanup for a new PR:

- Don't use `()` around `if` statements - While I can't see a recommendation in PEP08 this [1] question does say: [PEP08] _makes no mention of the second (parenthesized) style, at all; it makes the assumption that no-one would use it._ 

- Internal methods need to start with `_` to indicate private

- `get_appconfig_details` name has the potential to be confusing , since it's not used to get an app config. 

- Need to ensure testing handles setting an empty property set correctly for changing an app config, (matching the existing command).

PEP08 - https://www.python.org/dev/peps/pep-0008/
[1] - https://stackoverflow.com/questions/30549554/python-coding-style-parentheses-in-ifs-loops-etc

com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/streamtool.py
```diff
@@ -91,20 +92,170 @@ def _lsjobs(instance, cmd_args):
...
+
+# utility function for mk-appconfig and ch-appconfig
+# Create the config properties
+def create_appconfig_props(config_props, prop_list):
+    # Iterate through list of properties, check if correct format, convert each to name/value pairs add to config_props dict
+    # Ex. good proplist is ['name1=value1', 'name2=value2']
+    # Ex. bad proplist is ['name1=valu=e1', 'name2=value2']
+    for prop in prop_list:
+        name_value_pair = prop.split("=")
+        if (len(name_value_pair) is not 2):
```
ddebrunner: Using `is not` for numerical comparison is not typical, use `!=`.

PR #2154, opened 2019-07-03
fixed misc bugs

Karsten12:
- [x] Test should validate the return from main is zero for success, non-zero for errors
- [x] rmappconfig returns 1 when it fails to delete an non-existent app config
- [x] passing --User to mkappconfig causes nothing to be created with no output. Printed error (server side fix tbd)

### DIFF
--- a/test/python/scripts/test_st_appconfig.py
+++ b/test/python/scripts/test_st_appconfig.py
@@ -38,8 +38,10 @@ def captured_output():
 
 class TestAppconfig(unittest.TestCase):
 
-    instance = "zen-edge-sample-icp1-blitz-env"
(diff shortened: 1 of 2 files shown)


test/python/scripts/test_st_appconfig.py
```diff
@@ -38,8 +38,10 @@ def captured_output():
 
 class TestAppconfig(unittest.TestCase):
 
-    instance = "zen-edge-sample-icp1-blitz-env"
+    instance = os.environ['STREAMS_INSTANCE_ID']
```
ddebrunner: These fields should be instance fields, not class fields. Set them in a `setUp` method.

PR #2170, opened 2019-07-16
ICP4D authentication improvements.

ddebrunner:
Always use ICP4DAuthenticator in Java for ICPD requests, rather that just bearer authentication with the service token.

Support using an ICPD `Instance` as `ContextProperties.STREAMS_INSTANCE` with distributed remote build and submit.


ddebrunner:
Tested with Cloud Pak for Data.

PR #2188, opened 2019-07-23
--long opt in lsjobs

Karsten12:
- implemented --long opt in lsjobs command, added tests, modified docs accordingly
- All test cases still pass


com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/streamtool.py
```diff
@@ -246,6 +249,13 @@ def _lsjobs(instance, cmd_args, rc):
         h_length[5] = max(len(job.name), h_length[5])
         h_length[6] = max(len(jobGroup), h_length[6])
 
+        # If --long option, add productVersion to the job data
+        if cmd_args.long:
+            prod_version = job.json_rep['productVersion']
```
ddebrunner: Can't you just use `job.productVersion` ?

Karsten12: Fixed in latest commit

PR #2203, opened 2019-07-29
Step 2: Check dependencies version is valid 

Karsten12:


ddebrunner:
@Karsten12 In a follow on PR work needs to be done to support adependency version, which at least with a simple version: e.g. 1.2.3.

Spec is here: https://www.ibm.com/support/knowledgecenter/SSCRJU_4.3.0/com.ibm.streams.dev.doc/doc/toolkitinformationmodelfile.html

PR #2211, opened 2019-07-31
Step 3: check if dependency w/ correct version already on build-server

Karsten12:
- Checks whether each dependency w/ correct version is on build-server
- if yes, doesn't do anything
- if no, checks locally for dependency w/ correct version (and if yes, uploads it)

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
@@ -26,8 +26,6 @@ def main(args=None):
     cmd_args = _parse_args(args)
     topo = _create_topo(cmd_args)
 
-    # print(cmd_args.disable_ssl_verify)
-    # exit()
     # Get dependencies for app, if any at all
     dependencies = _parse_dependencies()
     # if dependencies and if -t arg, find & add local toolkits
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
@@ -74,6 +74,7 @@ def _add_local_toolkits(toolkit_paths, dependencies, topo):
             continue
         # Dependency w/ correct version not in buildserver, check locally
         else:
+            # print("Dependency {} not on buildserver, check locally".format(dependency_name))
             toolkit = _get_local_toolkit(dependency_name, dependency_version, local_toolkits)
             if toolkit:
                 # Dependency w/ correct version exists locally, add it
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
@@ -67,17 +67,17 @@ def _add_local_toolkits(toolkit_paths, dependencies, topo):
 
     local_toolkits = _get_all_local_toolkits(toolkit_paths)
 
-    # Iterate through all dependencies (toolkits) needed to build sab file, and check if the dependencies already exist on the build server
+    # Iterate through all dependencies (toolkits) needed to build sab file, and check if the dependency already exist on the build server
     for dependency_name, dependency_version in dependencies.items():
         if _check_toolkit_on_buildserver(dependency_name, dependency_version, sc):
-            # Dependency w/ correct version is already in buildserver, go on to next dependency
+            # Dependency w/ correct version is already on buildserver, thus don't need to do anything
             continue
         # Dependency w/ correct version not in buildserver, check locally
         else:
             toolkit = _get_local_toolkit(dependency_name, dependency_version, local_toolkits)
             if toolkit:
+                # Dependency w/ correct version exists locally, add it
                 add_toolkit(topo, toolkit)
-                continue
 
 def _get_all_local_toolkits(toolkit_paths):
     """ A helper function that given the local toolkit paths, creates and returns a list of local toolkit objects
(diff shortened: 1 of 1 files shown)


com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
```diff
@@ -51,45 +52,95 @@ def _parse_dependencies():
...
-    toolkit_name = name.text
-    toolkit_version = version.text
-    return _LocalToolkit(toolkit_name, toolkit_version, path)
+    Returns:
+        _LocalToolkit -- The matching local toolkit, or None if it doesn't exist
+    """
+    # Get all local toolkits that have the same name as our dependency
+    temp = [x for x in local_toolkits if x.name == dependency_name]
+    # For each toolkit that matches the name
+    for x in temp:
```
ddebrunner: As a style item `for x in temp` is hard for others to read as it's not clear what `x` or `temp` is. Useful to have variables named for what they represent.

com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
```diff
@@ -51,45 +52,95 @@ def _parse_dependencies():
...
 
     Arguments:
-        toolkits {List} -- A list of local toolkit directories, each could be a toolkit directory (has info.xml directly inside),
+        toolkit_paths {List} -- A list of local toolkit directories, each could be a toolkit directory (has info.xml directly inside),
                 or a directory consisting of toolkits (no info.xml directly inside)
-        dependencies {Dictionary} -- A dictionary where the key is the names of the dependencies, and the value is its version.
+        dependencies {Dictionary} -- A dictionary consisting of elements where the key is the name of the dependency, and the value is its version.
         topo {topology Object} -- [description]
     """
+    sc = BuildService.of_endpoint(verify=False)
```
ddebrunner: `sc` as a local variable name doesn't intuitively imply _build service_ and given this tool is **sc** the local variable could have a more appropriate name.

ddebrunner: At some point we can't hard-code `verify=false`, instead we need a option like the one added to `streamsx-streamtool`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
```diff
@@ -51,45 +52,95 @@ def _parse_dependencies():
...
+
+    local_toolkits = _get_all_local_toolkits(toolkit_paths)
+
+    # Iterate through all dependencies (toolkits) needed to build sab file, and check if the dependency already exist on the build server
+    for dependency_name, dependency_version in dependencies.items():
+        if _check_toolkit_on_buildserver(dependency_name, dependency_version, sc):
```
ddebrunner: This implementation results in a REST request on every dependent toolkit which will 99.9999% of the time return the same info, so can the rest request be done once, and then only if there are dependent toolkits.

ddebrunner:
This code looks basically correct but the end-goal is to match what `sc` does today, and this differs in that a matching toolkit is selected based upon the **highest version** that maps the dependency.

For example, if the dependency is `[1.0.0,2.0.0)` then:

 * with local toolkits 1.1.0, 1.1.2 and remote 1.2.3 then remote 1.2.3 should be used
 * with local toolkits 1.1.0, 1.3.5 and remote 1.2.3 then local 1.3.5 should be used

Looks like the existing and PR code (for local & remote) selects the first matching version, local or remote. Just wondering if it makes sense to add this now or as a subsequent PR?

Karsten12:
> This code looks basically correct but the end-goal is to match what `sc` does today, and this differs in that a matching toolkit is selected based upon the **highest version** that maps the dependency.
> 
> For example, if the dependency is `[1.0.0,2.0.0)` then:
> 
> * with local toolkits 1.1.0, 1.1.2 and remote 1.2.3 then remote 1.2.3 should be used
> * with local toolkits 1.1.0, 1.3.5 and remote 1.2.3 then local 1.3.5 should be used
> 
> Looks like the existing and PR code (for local & remote) selects the first matching version, local or remote. Just wondering if it makes sense to add this now or as a subsequent PR?

Will implement in next PR

com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
```diff
@@ -51,45 +52,98 @@ def _parse_dependencies():
...
+
+        # All the toolkits that match dependency_name regardless of version
+        matching_toolkits = [toolkit for toolkit in all_toolkits if toolkit.name == dependency_name]
+
+        if not matching_toolkits:
+            # No toolkits exists on the remote buildserver or locally that can satsify this dependency, continue and let SPL compiler throw error
+            continue
+
+        # Toolkit with the highest verion that still satisfies the dependency_version
+        latest_compatible_toolkit = matching_toolkits[0]
```
ddebrunner: Doesn't `latest_compatible_toolkit` need to start out as `None`?

`matching_toolkits[0]` is a match on name only, it may not satisfy the dependency, so if it is `5.0.0` but the dependency is `[1.0.0,2.0.0]` then it will be selected.

com.ibm.streamsx.topology/opt/python/packages/streamsx/scripts/sc.py
```diff
@@ -51,45 +52,98 @@ def _parse_dependencies():
...
+                # if it satisfies, compare it against the highest version we have seen thus far
+                latest_version_so_far = version.parse(latest_compatible_toolkit.version)
+                curr_tk_version = version.parse(tk.version)
+
+                # if it is a later version, update our toolkit
+                if curr_tk_version > latest_version_so_far:
+                    latest_compatible_toolkit = tk
+
+        # Check if latest_compatible_toolkit is local, bc we then need to add it
+        if latest_compatible_toolkit in local_toolkits:
```
ddebrunner: How about `isinstance(latest_compatible_toolkit, _LocalToolkit)` ?

PR #2214, opened 2019-08-01
Fixed bug on Lsjobs and lsapppconfig tests

Karsten12:
- Tests initially assumed no prior running/existing jobs and appconfigs, fixed in this commit

### DIFF
--- a/test/python/scripts/test_st_appconfig.py
+++ b/test/python/scripts/test_st_appconfig.py
@@ -401,7 +401,7 @@ class TestAppconfig(unittest.TestCase):
             if self.name in line:
                 config = line
 
-        # If we can't find our jobID in the output, should fail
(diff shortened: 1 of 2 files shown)


ddebrunner:
@Karsten12 Is it possible for the `job.id in line` tests to be more specific? Job ids are simple numbers and it seems there is a less than zero chance that the wrong line will be selected, e.g. with job id of 2 there's a good chance any line will match due to the a 2 being in a date/timestamp.

Karsten12:
> @Karsten12 Is it possible for the `job.id in line` tests to be more specific? Job ids are simple numbers and it seems there is a less than zero chance that the wrong line will be selected, e.g. with job id of 2 there's a good chance any line will match due to the a 2 being in a date/timestamp.

Fixed in commit  36dc39b by also checking for name (in addition to id)

PR #2230, opened 2019-08-11
Fixes for SC tests

Karsten12:
- Generate random toolkit names via script (auto called by test_sc.py) -> Generates toolkit’s info.xml and toolkit.xml from info.xml.tmpl and toolkit.xml.tmpl template files, and generates each app's info.xml from info.xml.tmpl template file to use the randomly generated toolkit name
    - test_sc.py also calls another script that deletes these generated files (each toolkit's info.xml and toolkit.xml, and each app's info.xml) after tests are finished
- Switched from packaging to pkg_resources.parse_version(__version__) for toolkit version checking
- Moved deletion of sab to teardown method in case of test failure 
- Simplified name of apps from com.example.test_app_1 to test_app_1
- Fixed bug where test toolkits fails to upload


test/python/scripts/toolkits/delete_test_sc_files.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Script used/called by test_sc.py to delete the info.xml and toolkit.xml files generated for the tests
+# Also updates the info.xml and toolkit.xml files with random toolkit names, so different users can run tests concurrently.
```
ddebrunner: copy/paste comment left in.

test/python/scripts/toolkits/delete_test_sc_files.sh
```diff
@@ -0,0 +1,22 @@
...
+# Do NOT call this script from the cmd/terminal
+
+# Delete info.xml files for the toolkits located in /scripts/toolkits/
+cd toolkits
+for i in $(find . -name info.xml); do
+    rm $i
+done
+
```
ddebrunner: FYI - `xargs` is typically used with `find`.

`find toolkits -name info.xml | xargs rm`

test/python/scripts/toolkits/delete_test_sc_files.sh
Karsten12: Resolved in #2232

test/python/scripts/toolkits/delete_test_sc_files.sh
Karsten12: Resolved in #2232

PR #2253, opened 2019-09-09
Enable JWT authentication for stand-alone configurations.

EricMichaelJohnson:


ddebrunner:
@EricMichaelJohnson What is an example value of `STREAMS_REST_URL` or `rest_endpoint`?

Is it the URL of the service (e.g. https://some_host:port) or the traditional Streams url (e.g. https://some_host:port/streams/rest/resources)?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -1636,50 +1682,67 @@ def _root_from_endpoint(endpoint):
         return resource_url, name
 
     @staticmethod
-    def of_endpoint(endpoint=None, service_name=None, username=None, password=None, verify=None):
+    def of_endpoint(endpoint=None, service_name=None, username=None, password=None, verify=None, rest_endpoint=None, security_endpoint=None):
```
ddebrunner: Is it possible to just have `endpoint` rather than `endpoint` and `rest_endpoint`.

Could the presence of `security_endpoint` indicate a stand-alone configuration?

EricMichaelJohnson:
An example rest endpoint is 'https://10.6.33.26:32162/streams/rest/resources'

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
EricMichaelJohnson: It would be possible to remove `rest_endpoint` and use `endpoint` instead.  I think it may be less confusing to have two different parameters, rather than a single parameter that is specified differently depending on whether it is a stand-alone or integrated configuration.

EricMichaelJohnson: However, currently `service_name` must be specified.  In a stand-alone configuration, is there always exactly one instance?  If so, we don't actually need to know the instance name, otherwise we do.

ddebrunner:
> An example rest endpoint is 'https://10.6.33.26:32162/streams/rest/resources'

We need to change that as in the k8s world, it's the sws server that's being exposed and it just has the path `https://10.6.33.26:32162`. Internally we can add `/streams/rest/resources` but the end-user doesn't see that, only the defined service endpoint.

And then by definition there is a single Streams instance per service endpoint, so `service_name` is not needed & should not be supported for a stand-alone configuration.

EricMichaelJohnson:
I have made two changes: in stand-alone, build and rest URLs do not have to end in '/streams/rest/resources/, and service_name is not required.

EricMichaelJohnson:
Note that the security service URL still must be fully specified.  I don't think there is a fixed path for that.

EricMichaelJohnson:
An example security service url: `https://10.6.176.19:32412/streams/rest/security/realms/test-realm/accesstokens` . I think that the realm (in this case `test-realm`) is not fixed and the user will need to either the full url or the realm.  Currently the full url must be provided.

ddebrunner:
Merged so incremental work can be done in `feature/stand-auth` as understanding of the setup clarifies.

PR #2260, opened 2019-09-17
Add support to Instance.ofEndpoint() for standalone configuration

johnmac-ibm:
This adds a `StandaloneAuthenticator` similar to `ICP4DAuthenticator`, and modifies `Instance.ofEndpoint()` to construct the appropriate identifier.

Note further changes will be required to support the equivalend of the CP4D functionality that allows the `Instance` to be used at `submit` since it does not yet have any way to discover the build URL, and changes to the submit build service code that currently expect only `ICP4DAuthenticator` will need to be updated to handle both types.

### DIFF
--- a/java/src/com/ibm/streamsx/rest/Instance.java
+++ b/java/src/com/ibm/streamsx/rest/Instance.java
@@ -191,7 +191,7 @@ public class Instance extends Element {
         if (possible_integ && name == null)
             name = System.getenv(Util.STREAMS_INSTANCE_ID);
 
-        StreamsConnection conn = null;
+        StreamsConnection conn;
         Instance instance = null;
         if (name != null) {
             // Integrated configuration
--- a/java/src/com/ibm/streamsx/rest/Instance.java
+++ b/java/src/com/ibm/streamsx/rest/Instance.java
@@ -10,15 +10,18 @@ import static java.util.Objects.requireNonNull;
 import java.io.File;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
 
 import com.google.gson.JsonObject;
 import com.google.gson.annotations.Expose;
 import com.ibm.streamsx.rest.internal.ICP4DAuthenticator;
 import com.ibm.streamsx.rest.internal.RestUtils;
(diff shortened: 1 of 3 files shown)


java/src/com/ibm/streamsx/rest/internal/StandaloneAuthenticator.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final String STREAMS_REST_RESOURCES = "/streams/rest/resources";
+    private static final String RESOURCES = "resources";
+    private static final String ACCESS_TOKENS = "accessTokens";
+    private static final String AUDIENCE_STREAMS = "{\"audience\":[\"streams\"]}";
+    private static final String EXPIRE_TIME = "expireTime";
+    private static final String ACCESS_TOKEN = "accessToken";
+    private static long MS = 1000;
+
+    public static StandaloneAuthenticator of(String endpoint, String userName, String password) {
+        if (endpoint == null) {
```
ddebrunner: Seems strange that this internal class is checking for environment variables. Won't the value(s) be always set by the `ofEndpoint` method?

java/src/com/ibm/streamsx/rest/Instance.java
```diff
@@ -132,29 +176,75 @@ public static Instance ofEndpoint(String endpoint, String name, String userName,
...
+                endpoint = System.getenv(Util.STREAMS_REST_URL);
+                if (endpoint == null) {
+                    return null;
+                }
+            }
+        }
+        if (possible_integ && name == null)
+            name = System.getenv(Util.STREAMS_INSTANCE_ID);
+
+        StreamsConnection conn = null;
```
ddebrunner: Don't initialize to `null` - the value is set in both paths and not setting it to `null` would correctly cause a compile error if it was used before set.

java/src/com/ibm/streamsx/rest/internal/StandaloneAuthenticator.java
johnmac-ibm: In the current usage, yes. I noticed that `ICP4DAutenticator.of()` was checking env vars because in the build / submit path it sometimes gets called with `null` for various params, and I wasn't sure if standalone would end up doing something similar.

I'm fine with removing those checks though and worrying about the build / submit path later if you like.

johnmac-ibm:
@ddebrunner  I was able to pass the CP4D tests using another instance. Let me know if you would like me to remove the environment var handling mentioned above, or any other changes you'd like before merging.

PR #2299, opened 2019-10-25
Added functionality to delete toolkits, includes tests 

Karsten12:
Functionality to delete toolkits off a build server using any of the following 3 ways:

1. Delete 1 toolkit with a given id 
2. Delete all toolkits (regardless of version) with a given name
3. Delete all toolkits (regardless of version) where the name matches a given regex

Uses the same random test toolkits from sc tests


ddebrunner:
Can the command be `rmtoolkit` to be consistent with other stream tool commands.

ddebrunner:
@Karsten12 Can the streamtool documentation also be updated.

PR #2306, opened 2019-11-08
Use provided capabilities to get specific toolkits

ddebrunner:
Adds `name` to `BuildService.get_toolkits` just as name is supported in existing rest calls that return a collection.

Use the `BuildService` method to filter the toolkits, where possible, including using `get_toolkit` when requesting to delete by name.


ddebrunner:
Passes `test_st_rmtoolkit.py`

PR #2315, opened 2019-11-14
Get Streamsx-sc to work with .splmm files

Karsten12:
- Includes tests for example .splmm (test_simple_4 and test_simple_5, test_simple_6)

- All streamsx-sc test cases pass after running `ant clean` and `ant`

### DIFF
--- a/java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
+++ b/java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
@@ -104,6 +104,8 @@ public class ZippedToolkitRemoteContext extends ToolkitRemoteContext {
         
         JsonObject graph = graph(submission);
         
+        // System.out.println(graph.toString());
+
         if (mainCompositeKind == null) {
             String namespace = splAppNamespace(graph);
             String name = splAppName(graph);
@@ -128,13 +130,17 @@ public class ZippedToolkitRemoteContext extends ToolkitRemoteContext {
         Path manifestTmp = Files.createTempFile("manifest_tk", ".txt");
         Path mainCompTmp = Files.createTempFile("main_composite", ".txt");
         Path scOptsTmp = Files.createTempFile("sc_opts", ".txt");
+        Path splmmOptsTmp = Files.createTempFile("splmm_opts", ".txt");
+        Path mainCompDirTmp = Files.createTempFile("mainCompDir", ".txt");
         
+        // String[] mainCompDir1 = {""};
+        // Map<Path,String> mainCompDir = new HashMap<>();
+
         // tkManifest is the list of toolkits contained in the archive
         try (PrintWriter tkManifest = new PrintWriter(manifestTmp.toFile(), "UTF-8")) {
             if (tkName != null)
                 tkManifest.println(tkName);
             tkManifest.println(topologyToolkitName);
-            
             JsonObject configSpl = object(graph, CONFIG, "spl");
             if (configSpl != null) {
                 objectArray(configSpl, "toolkits",
@@ -144,11 +150,31 @@ public class ZippedToolkitRemoteContext extends ToolkitRemoteContext {
                                 String tkRootName = tkRoot.getName();
                                 tkManifest.println(tkRootName);
                                 toolkits.put(tkRoot.toPath(), tkRootName);
+
+                                // if (deploy(submission).has(ContextProperties._SPLMM_OPTIONS)) {
+                                //     // if (mainCompDir1[0] == "") {
+                                //     //     mainCompDir1[0] = tkRootName;
+                                //     //     // System.out.println("------------- " + tkRoot.toPath().toString() + " -------------");
+                                //     //     // System.out.println("------------- " + tkRoot.getName() + " -------------");
+                                //     // }
+                                //     if (mainCompDir.isEmpty()) {
+                                //         mainCompDir.put(tkRoot.toPath(), tkRootName);
+                                //         // System.out.println("------------- " + tkRoot.toPath().toString() + " -------------");
+                                //         // System.out.println("------------- " + tkRoot.getName() + " -------------");
+                                //     }
+                                // } else {
+                                //     tkManifest.println(tkRootName);
+                                //     toolkits.put(tkRoot.toPath(), tkRootName);
+                                //     System.out.println(" TKS ------------- " + tkRoot.toPath().toString() + " -------------");
+                                //     System.out.println(" TKS------------- " + tkRoot.getName() + " -------------");
+                                // }
                             }
                             }
                         );
             }
         }
+
+        // String mainCompDir = mainCompDir1[0];
         
         // mainComposite is a string of the namespace and the main composite.
         // This is used by the Makefile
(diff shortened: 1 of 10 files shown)


java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
```diff
@@ -117,6 +117,9 @@ private static Path pack(final File appTkRoot, JsonObject submission, String tkN
         
         // Paths to completely copy into the code archive
         Map<Path, String> paths = new HashMap<>();
+
+        Map<Path, String> SPLMM_toolkits = new HashMap<>();
+        String[] splmm_dir = {""};
```
ddebrunner: Why is this an array?

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
```diff
@@ -128,22 +131,33 @@ private static Path pack(final File appTkRoot, JsonObject submission, String tkN
...
                         tk -> {
                             if (tk.has("root")) {
                                 File tkRoot = new File(jstring(tk, "root"));
                                 String tkRootName = tkRoot.getName();
-                                tkManifest.println(tkRootName);
-                                toolkits.put(tkRoot.toPath(), tkRootName);
+                                // If SPLMM app, need to seperate this toolkit to run the command
+                                // 'spl-make-toolkit -i <SPLMM_APP_DIR> `cat splmm_opts.txt`'
+                                // Bc SPLMM app requires arguments, so can't group with other toolkits
+                                if (deploy(submission).has(ContextProperties._SPLMM_OPTIONS)) {
```
ddebrunner: Won't this miss toolkits if SPLMM is set? E.g. with toolkits A,B,C then A will be the SPLMM toolkits and the other two will be incorrectly ignored?

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
Karsten12: For line 121, I mimicked how toolkits were created using a map, and for line 122, I couldn't use a regular string for the splmm dir, bc where I set the value (in the lambda function on line 154) if I use a regular string i get the java error 'local variables referenced from a lambda expression must be final or effectively final', and one of the answers stated that a String array is a workaround.

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
Karsten12: I separate the main composite toolkit from the rest of the toolkits if it is an SPLMM app requiring arguments bc if I don't separate it, the makefile will error on this line '${STREAMS_INSTALL}/bin/spl-make-toolkit -i $${toolkit}; \'. 

ContextProperties._SPLMM_OPTIONS is only set in streamsx-sc, where if the main composite is determined to be an SPLMM app, that first toolkit (the splmm app) will be filtered out, while the rest of the toolkits will be processed by the else case as normal. Is this the correct behavior or should I be doing processing it some other way?

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
ddebrunner: Ok - makes sense but the typical value for not set would be `null`, not an empty string.

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
ddebrunner: How will the rest of the toolkits in that case be processed by the else?

 If the code archive has three toolkits A,B,C and _SPLMM_OPTIONS is set then A will go through lines 153-156, then B,C will fail the check at line 153 and not be added to `toolkits`.

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
Karsten12: I will change it to null and re-run the tests

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
Karsten12: Oh I see, sorry about that. Will change it and re-run the tests

Karsten12: Fixed in latest commit

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
Karsten12: Fixed in latest commit

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
Karsten12: Fixed in latest commit

ddebrunner:
@Karsten12 Build is failing - I think due to using `SPLMM_toolkits` instead of `splmm_dir`.

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
```diff
@@ -180,33 +196,68 @@ private static Path pack(final File appTkRoot, JsonObject submission, String tkN
...
+            if (!splmmOptions.isEmpty()) {
+                try (PrintWriter splmmOptsW = new PrintWriter(splmmOptsTmp.toFile(), "UTF-8")) {
+                    for (String splmmOpt : splmmOptions) {
+                        splmmOptsW.print(splmmOpt);
+                        splmmOptsW.print(" ");
+                    }
+                }
+            }
+
+            if (splmm_dir[0] != "") {
```
ddebrunner: This needs to be `!= null` right?

Karsten12: Oh, I keep forgetting to run `ant`, sorry, will fix

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
```diff
@@ -128,22 +131,36 @@ private static Path pack(final File appTkRoot, JsonObject submission, String tkN
...
                             if (tk.has("root")) {
                                 File tkRoot = new File(jstring(tk, "root"));
                                 String tkRootName = tkRoot.getName();
-                                tkManifest.println(tkRootName);
-                                toolkits.put(tkRoot.toPath(), tkRootName);
+                                // If SPLMM app, need to seperate this toolkit to run the command
+                                // 'spl-make-toolkit -i <SPLMM_APP_DIR> `cat splmm_opts.txt`'
+                                // Bc SPLMM app requires arguments, so can't group with other toolkits
+                                if (deploy(submission).has(ContextProperties._SPLMM_OPTIONS)) {
+                                    if (SPLMM_toolkits.isEmpty()) {
```
ddebrunner: Is there any reason to split the SPLMM toolkit into its own map, rather than just including it in `toolkits`. It seems the same code is executed for both.

Karsten12: Yes, it goes back to the case I spoke about above. If I don't separate it, the makefile will error on this line `${STREAMS_INSTALL}/bin/spl-make-toolkit -i $${toolkit}; \` since the SPLMM toolkit args are missing. An alternative that worked before was to edit that line to be `${STREAMS_INSTALL}/bin/spl-make-toolkit -i $${toolkit} `cat splmm_opts.txt`; \` but I wasn't confident that the SPLMM opts would not interfere with the other toolkits.

Karsten12:
All test cases pass, although sometimes 1 or 2 seems to fail

java/src/com/ibm/streamsx/topology/internal/context/remote/ZippedToolkitRemoteContext.java
ddebrunner: Isn't that command driven by `tkManifest`, not `toolkits`?

Isn't `toolkits` (from line 125) just used as the list of toolkits to copy into the code archive, which is exactly what `SPLMM_toolkits` (line 121) is doing?

Karsten12: I see what you're saying. I will try that and re-run the tests

Karsten12: Removed SPLMM toolkit map and included SPLMM toolkit in toolkits map, in bcb9110

Karsten12:
All test still cases pass, although sometimes 1 or 2 seems to fail

PR #2317, opened 2019-12-06
Added updateoperator function to streamtool

Karsten12:
- All test cases pass

Minor note: 
the test_parallelRegionWidth checks that the passed in parallel region width (the --parallelRegionWidth arg) overrides the one in the JCO, and I've confirmed it works (via printing out the JCO after the override), but this print statement has been removed, since its used for debugging only. Thus this test is essentially useless.
When the 1Q20 fix is implemented, and the proper JSON results of the updateoperator function is returned,  I will have to update this test to check for the correct parallel region width


com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -2780,6 +2791,24 @@ def _cancel_job(self, job, force):
 
         return False
 
+    def _update_operators(self, job, job_config):
+        self.rest_client._block_ssl_warn()
+
+        job_options = job_config.as_overlays() if job_config else {}
```
ddebrunner: Does an empty job_options make sense?

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -2780,6 +2791,24 @@ def _cancel_job(self, job, force):
 
         return False
 
+    def _update_operators(self, job, job_config):
+        self.rest_client._block_ssl_warn()
```
ddebrunner: Should be removed before final merge.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
```diff
@@ -2780,6 +2791,24 @@ def _cancel_job(self, job, force):
...
+
+        job_options = job_config.as_overlays() if job_config else {}
+
+        # Update the job operators using the job id
+        update_url = job.instance + '/jobs/' + job.id
+        res = self.rest_client.session.patch(update_url,
+                headers = {'Accept' : 'application/json'},
+                json = {'jobConfigurationOverlay':job_options, 'preview':False},
```
ddebrunner: `preview` should be passed in, defaulting to `False`.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
ddebrunner: This call is needed, sorry I thought this was always disabling SSL.

com.ibm.streamsx.topology/opt/python/packages/streamsx/rest_primitives.py
Karsten12: Removed in 6b53625

PR #2321, opened 2020-01-07
Support for composite transforms in source, map & for_each

ddebrunner:


ddebrunner:
Initial support for #2322

Issue #2360, opened 2020-03-10
Cannot submit to public cloud service using credentials only


natashadsilva:
According to 
https://streamsxtopology.readthedocs.io/en/latest/streamsx.topology.context.html#streamsx.topology.context.ConfigParams.SERVICE_DEFINITION

If you submit to the streaming analytics service, you should be able to only specify credentials when you configure job submission, you get this error:

```
Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <"> at index 1: {"streaming-analytics": [{"credentials": {"abc": "def"}, "name": "service"}]}
at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:194)
at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:165)
at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:89)
at sun.nio.fs.WindowsPath.parse(WindowsPath.java:106)
at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:267)
at java.nio.file.Paths.get(Paths.java:95)
```
Attached a notebook sample.

I tried the second option of 
>A JSON object (dict) of the form: { "type": "streaming-analytics", "name": "service name", "credentials": {...} } with the service credentials as the value of the credentials key.

and also got the same error.

natashadsilva:
[PublicCloudSubmitFailure.zip](https://github.com/IBMStreams/streamsx.topology/files/4314982/PublicCloudSubmitFailure.zip)

natashadsilva:
@markheger

RolefH:
### Reproduction:
**OS running the jupyter-notebook:** linux
**Environment variables:**
unset STREAMING_ANALYTICS_SERVICE_NAME, VCAP_SERVICES, all STREAMS_*


### First option (The service credentials copied from the Service credentials page)
- does *not* work for me
- Notebook code:
```
SA_credentials = getpass.getpass('Streaming Analytics credentials:')
print(type(SA_credentials))    # must be "<class 'str'>"
cfg[ConfigParams.SERVICE_DEFINITION] = SA_credentials
```
Ends up in `TypeError`:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-6-05dffd248a5a> in <module>
----> 1 submission_result = submit_topology(topo)

<ipython-input-4-04db718cb996> in submit_topology(topo)
[...]

PR #2370, opened 2020-03-12
convert SAS credentials to JSON object if given as string, issue #2360

RolefH:
JSON credentials given as str (copy&paste directly from cloud service console) must be converted to a JSON object. This is done in the Base class of all submitters, and in `_vcap_from_service_definition()` where a VCAP for JAVA is created.


RolefH:
Before merging this PR, I want to wait for response from @natashadsilva to #2360 as it is not clear, what happened there.

PR #2388, opened 2020-03-17
tests updated

markheger:


build.xml
```diff
@@ -124,6 +124,9 @@
     <delete>
       <fileset dir="test/python/scripts/spl_app" includes="toolkit.xml"/>
     </delete>
+    <delete>
+      <fileset dir="test/python/scripts/sc_test_files" includes="test_sc_old_toolkit_name.txt"/>
```
Karsten12: Not too familiar with how the ant build process works, does this delete the directory `scripts/sc_test_files`, or just the file `test_sc_old_toolkit_name.txt` ?

markheger: just the file

PR #2389, opened 2020-03-18
test update

markheger:
resolve issue in test_st_rmtoolkit.py

### DIFF
--- a/test/python/scripts/test_st_rmtoolkit.py
+++ b/test/python/scripts/test_st_rmtoolkit.py
@@ -73,7 +73,7 @@ class Testrmtoolkit(unittest.TestCase):
             self.skipTest("Build REST API is not available")
         else:
             # delete all the test toolkits from the buildserver, in case they were left behind by a previous test failure.
-            file_name = (my_path / "test_sc_old_toolkit_name.txt").resolve()
+            file_name = (my_path / "sc_test_files/test_sc_old_toolkit_name.txt").resolve()
             old_toolkit_name = None
             if os.path.isfile(file_name):
                 with open(file_name) as f:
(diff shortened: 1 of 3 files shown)


test/python/scripts/test_st_rmtoolkit.py
```diff
@@ -92,10 +92,11 @@ def setUp(self):
             # Get and store the randomly generated toolkit name, thus in case of network failure, on next run on test suite,
             # can read old randomly generated toolkit name, and delete toolkits off buildserver
             # Ex. 'com.example.tmyuuwkpjittfjla.test_tk_2' -> 'com.example.tmyuuwkpjittfjla.'
-            tk1 = self._get_toolkit_objects([self.test_toolkit_paths[0]])[0]
-            self.random_name_variable = re.sub("test_tk_.", "", tk1.name)
-            with open((my_path / "test_sc_old_toolkit_name.txt").resolve(), "w") as file1:
-                file1.write(self.random_name_variable)
+            if len(self.test_toolkit_paths) > 0:
```
Karsten12: Hey @markheger, correct me if i'm wrong, but I believe this if check is not needed. After you fixed the path issue in 64c4f25 (sorry btw, I completely missed fixing this), the `_get_remote_test_tk_objects` function has this [line](https://github.com/markheger/streamsx.topology/blob/1f28198dd7f56fecc6fd7294db6cf978f9c6d064/test/python/scripts/test_st_rmtoolkit.py#L266) that will cause the tests to prematurely fail if the number of `remote_test_tk_objects` does not equal the number of `test_toolkit_objects` that we uploaded. So in each of the 3 tests, `self.test_toolkit_paths` should never be empty

markheger: yes, not needed, but doesn't disturb, no need to change again

PR #2394, opened 2020-03-20
get_view_obj() removed, resolves #942

### LINKED ISSUE #942: Remove get_view_obj() from rest_primitives.py
Should `get_view_obj()` be removed? 
It seems to be a leftover artifact from refactoring.

markheger:


markheger:
> `get_view_object` is unused in this Python package - safe to delete. The assumption is, that the function is not part of the public API as it has no documentation?

No sample or test case found, that uses this function

rest tests passed

PR #2449, opened 2020-05-13
baseImage query and selection added

RolefH:
The algorithm to select a base image is:
1. Look for a build pool of type **image**
2. chose the first one of these pools
3. query the base images under that build pool
4. chose the first base image that contains _conda_ or _python_ in the name or tag converted to lowercase
5. if no such base image can be found, simply chose the first of the returned JSON array


RolefH:
Don't know if it is worth to create a class _BuildPool_, similar to _BaseImage_ to get rid of the JSON handling when getting the base images.

PR #2460, opened 2020-05-14
Obtain version in release file name from info.xml

joergboe:
With this change the release file version is obtained from info.xml (but the full version including alpha 1.15.0.alpha)
There is no longer a need to change the build.xml all the way.
If you like this, you can pull it.
We could also include the short git has into the release file name?

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -254,11 +254,11 @@
        <ant dir="python" target="pythondoc"/>
   </target>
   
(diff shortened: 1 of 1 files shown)


RolefH:
No git commit hash, please. The releases are tagged.

RolefH:
Merging the PR now, and doing the required change by myself.

Issue #2447, opened 2020-05-12
Support invocation to build an application for the edge


markheger:
For the Python developer, we add two new submission **contexts**:
`EDGE`, and `EDGE_BUNDLE`.

* `EDGE_BUNDLE` is intended for debugging purpose.
* `EDGE` is supposed to create the image.

markheger:
```
from streamsx.topology.topology import Topology
from streamsx.topology.context import submit, ContextTypes
            
topo = Topology()
...
submission_result = submit(ContextTypes.EDGE, topo.graph, cfg)
```

PR #2463, opened 2020-05-19
query of base images added to BuildService

RolefH:
```
from streamsx.build import BuildService

bs = BuildService.of_endpoint(verify=False)
baseImages = bs.get_base_images()
print('# images = ' + str(len(baseImages)))
for i in baseImages:
    print(type(i))
    print(i.id)
    print(i.registry)
```


markheger:
edge feature #2447

PR #2553, opened 2020-08-24
notes added for punctuation mode

markheger:


com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -2294,6 +2328,15 @@ def print(self, tag=None, name=None, write_punctuations=None):
         If `tag` is not `None` then each tuple has "tag: " prepended
         to it before printing.
 
+        .. note::
+             Punctuation marks are in-band signals that are inserted between tuples in a stream. Window punctuations are inserted into a stream that are related to the semantics of the operator. One example is the :py:meth:`~Window.aggregate`, which inserts a window marker into the output stream after each aggregation. 
+             
+             There are two kinds of punctuation markers, that are written to stdout when `write_punctuations` is set to `True`:
```
ghost: Suggestion: There are two kinds of punctuation markers, **which** are written to stdout ...

com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/topology.py
```diff
@@ -2294,6 +2328,15 @@ def print(self, tag=None, name=None, write_punctuations=None):
         If `tag` is not `None` then each tuple has "tag: " prepended
         to it before printing.
 
+        .. note::
+             Punctuation marks are in-band signals that are inserted between tuples in a stream. Window punctuations are inserted into a stream that are related to the semantics of the operator. One example is the :py:meth:`~Window.aggregate`, which inserts a window marker into the output stream after each aggregation.
```
ghost: _Window punctuations are inserted into a stream that are related to the semantics of the operator._
(Multiple places)

**Better (in my opinion):** If sources or stream transforms insert window markers at all, and when they insert them depends on the source or the semantic of the stream transformation.

The term _operator_ originates in SPL, I'm not sure if this term is established terminology in the topology API.

PR #2586, opened 2020-09-29
mvn release test

markheger:
Test added to download the topology.jar from Maven and build a Java Topology

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -168,6 +168,15 @@ cd test/release
 ant
 ```
 
+### Perform Java Topology quick test with release from Maven Central
+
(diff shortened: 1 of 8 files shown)


TESTING.md
```diff
@@ -168,6 +168,15 @@ cd test/release
 ant
 ```
 
+### Perform Java Topology quick test with release from Maven Central
+
+To perform a quick test (download topology.jar from Maven and build bundle from Java Topology app), run the following:
+
+```
+cd test/release
```
ghost, line 177: Add a line to set the environment variable TOPOLOGY_VERSION, otherwise it is always 1.16.0 unless the Makefile is edited.

export TOPOLOGY_VERSION=<tested version, for example 1.16.3>

PR #2608, opened 2020-11-19
Pydoc fixes

### LINKED ISSUE #2607: Documentation: Syntax error in Stream.split() example
https://streamsxtopology.readthedocs.io/en/v1.17.1/streamsx.topology.topology.html#streamsx.topology.topology.Stream.split

See `msgs` and `msg`:
```
msgs = topo.source(ReadMessages())
SEVS = {'H':0, 'M':1, 'L':2}
severities = msg.split(3, lambda SEVS.get(msg.get('SEV'), -1),
    names=['high','medium','low'], name='SeveritySplit')

high_severity = severities.high
high_severity.for_each(SendAlert())

medium_severity = severities.medium
medium_severity.for_each(LogMessage())

low_severity = severities.low
low_severity.for_each(Archive())
```

markheger:
resolves #2607 
fixes typos on streamsx.topology.state page

### DIFF
--- a/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/state.py
+++ b/com.ibm.streamsx.topology/opt/python/packages/streamsx/topology/state.py
@@ -38,9 +38,9 @@ Stateful callables
 Use of a class instance allows a transformation (for example :py:meth:`~streamsx.topology.topology.Stream.map`) to be stateful by maintaining state in instance
 attributes across invocations.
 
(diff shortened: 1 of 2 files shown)


ghost (approved):
When you correcting PyDoc, please correct also the doc for Stream.punctor(...):

_**replace** (bool) – If the value is True, then in case func(t) returns True the window punctuation will be generated and the tuple is discarded (not forwarded). The parameter before is not ignored in this case._

IMO, the parameter **before** _is ignored_ when **replace** is True. I haven't looked into the code, what is implemented, but  when the tuple is replaced by a window marker, the before parameter has no meaning.

markheger:
> not

"not" needs to be removed
The parameter before is ~~not~~ ignored in this case.

